PR #1, opened 2025-11-17
naber's stuff

alemreM:
wym cant automatically merge


CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
```diff
@@ -722,40 +723,64 @@ private void SetUpCastStanceParticles()
...
+            // apply to all heroes
+            var roster = mainParty.MemberRoster.GetTroopRoster();
+            for (int i = 0; i < roster.Count; i++)
             {
+                var element = roster[i];
+                if (!element.Character.IsHero)
+                    continue;
+
+                var hero = element.Character.HeroObject;
+                if (hero == null)
```
SlyDevil: Character.IsHero looks for a non-null HeroObject already, no?

party.GetMemberHeroes can return a list of heroes and performs the same checks if you want to use the existing extension

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentIngredientLootCampaignBehavior.cs
```diff
@@ -58,26 +56,60 @@ public override void SyncData(IDataStore dataStore)
...
+        }
+
+        mapEvent.GetBattleRewards(
+            PartyBase.MainParty,
+            out var _renown, out var _influence, out var _morale, out var _gold,
+            out var playerLootShare
+        );
 
-        foreach (var pair in _goodAmounts)
+        var targetRoster = PlayerEncounter.Current?.RosterToReceiveLootItems ?? PartyBase.MainParty.ItemRoster;
```
SlyDevil: I'd toss in an info message that tells the player that the items were added directly to their inventory here so they don't feel like they missed out on something if the encounter ends up null for whatever reason, but that they still received loot from it.

Or perhaps logging with TORCommon.Log so we have a message about when/how the first option could be null. Could be removed after beta testing so any performance concern is only present during testing.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentIngredientLootCampaignBehavior.cs
```diff
@@ -58,26 +56,60 @@ public override void SyncData(IDataStore dataStore)
...
+
+        foreach (var ingredientType in ingredientKeys)
         {
-            var amount = model.CalculateResultAmount(pair.Value, pair.Key, playerEarnedLootPercentage);
-            var item = TorEnchantingIngredients.GetItemObjectForIngredient(pair.Key);
-            itemRosterToReceive.Add(new ItemRosterElement(item, amount));
+            var factorSum = _goodAmounts[ingredientType];
```
SlyDevil: Can you rename _goodAmounts to _goodsFactors? I think the prior usage dates back to the original writing by zerca which may be different than what it's now used for.

CSharpSourceCode/CampaignMechanics/CustomResourceBehavior/TeefBehavior.cs
```diff
@@ -361,6 +313,32 @@ private void AddGoldForTeef(List<InquiryElement> inquiryElements)
...
+    {
+        long totalItemValue = 0;
+
+        foreach (var element in itemRoster)
+        {
+            var item = element.EquipmentElement.Item;
+            if (item == null)
+                continue;
+
+            if (item.StringId != null && item.StringId.StartsWith("tor_gs_"))
```
SlyDevil: Are there items that have ids starting with "tor_gs_" that aren't marked as Culture.aserai?

This can probably be checked more directly via (item.Culture == TORConstants.Cultures.GREENSKINS).

CSharpSourceCode/CampaignMechanics/CustomResourceBehavior/TeefBehavior.cs
```diff
@@ -151,63 +151,15 @@ void CreateTeefContainer(List<InquiryElement> inquiryElements)
...
-                    if (currentRoster.FindIndexOfElement(item.EquipmentElement) != -1)
-                    {
-                        continue;
-                    }
-                    difference.Add(item);
-                }
+            InventoryScreenHelper.OpenScreenAsReceiveItems(
+                donatedItems,
+                new TextObject("Give Items to the Big boss"),
+                () => OnItemsDiscarded(donatedItems));
```
SlyDevil, line 159: much cleaner, very nice

CSharpSourceCode/CampaignMechanics/CustomResources/WaaaghMeter/WaaaghBehavior.cs
```diff
@@ -49,39 +54,56 @@ private void CalculateWaaaghGainFromBattle(MapEvent mapEvent)
     {
         // Only apply to Greenskin players
         if (Hero.MainHero.Culture.StringId != TORConstants.Cultures.GREENSKIN) return;
-
-        if (!mapEvent.IsPlayerMapEvent) return;
+        if (mapEvent == null || !mapEvent.IsPlayerMapEvent) return;
 
         var playerWon = mapEvent.WinningSide == mapEvent.PlayerSide;
 
-        if (playerWon)
```
SlyDevil, line 78: Keep the playerWon check here.

If they lost, the delta can be set immediately without bothering to calculate the ratio and other information that will have no bearing.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -114,58 +116,129 @@ public List<string> GetUnlockedOakUpgradeCategory(string unlockedUpgradeCategory
...
+                    .SelectMany(r => r.ReligiousArtifacts)
+                    .Select(i => i.StringId)
+                    .ToHashSet() ?? new HashSet<string>();
+
+                var heroCulture = Hero.MainHero.Culture;
+
+                var rewardItems = comp.RewardItemIds
+                    .Select(id => MBObjectManager.Instance.GetObject<ItemObject>(id))
+                    .Where(i => i != null)
+                    .Where(i => _xmlItemIds.Contains(i.StringId))
```
SlyDevil: I'm not so sure about this check.

_xmlItemIds is filled via OnSessionLaunched which will trigger for both new and saved games. Does this event precede the save game being read and any additional item ids not present in the xml (ie. enchanted items generated previously and baked into the save) having already been added to the ObjectManager for its "Items" list?
Have you verified that enchanted items gained aren't present in the hash set?

If they are, then specificItems will include them and the player may be offered a magic item they already have (as items not in their inventory are periodically deleted).

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -114,58 +116,129 @@ public List<string> GetUnlockedOakUpgradeCategory(string unlockedUpgradeCategory
...
+                    .Where(i => !artifactIds.Contains(i.StringId))
+                    .ToList();
+
+                List<ItemObject> blueprintRewardItems = new();
+                if (rewardItems.Count > 0)
                 {
-                    itemIds = itemIds.TakeRandom(2).ToList();
-                }
+                    var rewardSelection = rewardItems;
```
SlyDevil, line 159: This logic creates a weird boundary condition when the raid component only contains 1 or 2 items in its list.

I'm also confused by the random selection, then filtering for the recipes to learn, rather than selecting from just the recipes with some chance scaled by rewardItems.Count.
Is the goal in separating out the recipes to dilute the equipment items from the drop list with the cultural items while maintaining the drop rates on the recipes?

Recipes can probably be filtered directly by checking the item's type is "Goods".

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -114,58 +116,129 @@ public List<string> GetUnlockedOakUpgradeCategory(string unlockedUpgradeCategory
...
-                var items = itemIds.Select(id => MBObjectManager.Instance.GetObject<ItemObject>(id)).ToList();
+                    blueprintRewardItems = rewardSelection
+                        .Where(i => i.StringId.StartsWith("tor_learn_"))
+                        .ToList();
+                }
 
+                var specificItems = comp.RewardItemIds
+                    .Select(id => MBObjectManager.Instance.GetObject<ItemObject>(id))
+                    .Where(i => i != null)
+                    .Where(i => _xmlItemIds.Contains(i.StringId))
```
SlyDevil: If the item exists in the object manager, and has been listed in the drop list, then it must necessarily exist in the hashset because that was created from the ObjectManager.

Rather than just null checking here, the drop lists should be verified on campaign initialization and listed items that don't yet exist in the weapon xmls should be logged to preemptively keep track of problems.

The TorItem check should be superfluous; if we've added native equipment to our drop lists, that's our problem to take care of at a different point rather than rechecking here.

Similarly, CraftedByPlayer and artifacts should not be added to the drop list by us.

The culture check is confusing. Much of the drop table will be filtered out by it.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -114,58 +116,129 @@ public List<string> GetUnlockedOakUpgradeCategory(string unlockedUpgradeCategory
...
+                    .Where(i => i.IsTorItem()) // no vanilla
+                    .Where(i => (i.IsWeapon() || i.IsArmor()))
```
SlyDevil: If someone wants a shield to drop from the raid, this will prevent that from ever dropping with no indication as to why despite the expectation of the drop list being the "intended items".

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -114,58 +116,129 @@ public List<string> GetUnlockedOakUpgradeCategory(string unlockedUpgradeCategory
...
+                    }
+                }
+
+                items = items
+                    .OrderBy(_ => MBRandom.RandomFloat)
+                    .ToList();
 
                 var model = (TORBattleRewardModel)Campaign.Current.Models.BattleRewardModel;
 
                 var newItems = new List<ItemObject>(items);
```
SlyDevil: Why not take the ~4 items from equipmentItems, modify them with traits, and add them directly to the final reward list and avoid having to copy book+equipment to items, items to newItems, filtering items to not add traits to the book, then adding the modified equipment piece to newItems and removing the unmodified version from newItems?

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -359,6 +443,10 @@ private void OnSessionLaunched(CampaignGameStarter starter)
         }
 
         CollectSettlementData();
+        if (_xmlItemIds == null)
+            _xmlItemIds = MBObjectManager.Instance.GetObjectTypeList<ItemObject>()
```
SlyDevil: Has this been checked for player crafted items to verify that no enchanted items have been loaded into the manager at this point?

CSharpSourceCode/HarmonyPatches/MedicineSurgeryChancePatch.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Collections.Generic;
+using System.Reflection.Emit;
+using HarmonyLib;
+using TaleWorlds.CampaignSystem.GameComponents;
+
+namespace TOR_Core.HarmonyPatches
+{
+    [HarmonyPatch(typeof(DefaultPartyHealingModel), "GetSurvivalChance")]
```
SlyDevil: Was there a reason to do this via a transpile rather than our existing override for that method?

SlyDevil (commented):
Looks good. A couple things that stand out.

Not sure why it picked up so many changes from other commits and bloated the files changed list.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentIngredientLootCampaignBehavior.cs
```diff
@@ -92,21 +92,30 @@ private void SetLootedIngredients(MapEvent mapEvent)
...
             if (amount <= 0) continue;
 
             var item = TorEnchantingIngredients.GetItemObjectForIngredient(ingredientType);
+            anyIngredientAdded = true;
             targetRoster.Add(new ItemRosterElement(item, amount));
         }
+        if (usedLootFallback && anyIngredientAdded)
+        {
+            InformationManager.DisplayMessage(new InformationMessage(
+                new TextObject("{=tor_crafting_ingredient_loot_fallback}Looted enchantment ingredients were added directly to your party inventory.").ToString()));
```
SlyDevil: str_tor_crafting_ingredient_loot_fallback

string id starts with "tor", localization id starts with "str_tor"

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -221,15 +249,24 @@ private void OnMissionEnded(IMission obj)
...
                 var model = (TORBattleRewardModel)Campaign.Current.Models.BattleRewardModel;
 
-                var newItems = new List<ItemObject>(items);
+                var unmodifiedItems = new List<ItemObject>(items.Count);
+                var enchantedItems = new List<ItemObject>();
+
                 foreach (var item in items)
                 {
-                    if (!item.IsWeapon() && (!item.IsArmor() || item.HasAnyTrait()))
+                    var canReceiveTraits = item.IsWeapon() || (item.IsArmor() && !item.HasAnyTrait());
```
SlyDevil: `(item.IsWeapon() || item.IsArmor()) && !item.HasAnyTrait()`

Current parentheses check for IsWeapon OR (IsArmor and the armor has no traits), but that feels off to me.

Looking at the current code before your changes and I'm just as confused by the original logic; why would a weapon that already has a trait have more added, but an armor with a trait can't have any more added.
`ItemTrait.IsValidFor` has a case to cover shield itemtype, and the incoming items were already filtered to weapon/armor/shield, so this bool can simply be `!item.HasAnyTrait()`.

For consistency between weapons and armor+shield, I find the simplest to simply exclude anything with a pre-existing trait from receiving more which is what I wrote above.
There's a potential to add parity by checking the difference between `item.GetTraits().Count` and traitCount, to then add the difference of traits to the item regardless of type, but I'd leave that out for now and ask others to weigh in on preferences for a future change.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -239,10 +276,12 @@ private void OnMissionEnded(IMission obj)
                     var name = model.GetNameModifierForTraits(traitCount);
                     var newItem = EnchantmentHelper.CreateEnchantedItem(item, ids, name + " " + item.Name, false);
 
-                    newItems.Add(newItem);
-                    newItems.Remove(item);
+                    enchantedItems.Add(newItem);
                 }
-                items = newItems;
+
+                unmodifiedItems.AddRange(enchantedItems);
```
SlyDevil, line 287: Add `blueprintRewardItems` as well; recipes are currently not added even if they were selected as a drop further up.

unmodifiedItems comes from equipmentItems which comes from selectedItems/cultureItems, but none of these prior lists had the recipes added to them and specifically filtered non-equipment out.

ModuleData/tor_strings.xml
```diff
@@ -2796,7 +2804,68 @@
...
+	<string id="tor_spelltrainer_greenskin_start_str" text="{=tor_spelltrainer_greenskin_start_str}You come to look into da fire?" />
+	<string id="tor_spelltrainer_greenskin_hub_str" text="{=tor_spelltrainer_greenskin_hub_str}Wot you want, git?" />
+	<string id="tor_spelltrainer_greenskin_open_book_str" text="{=tor_spelltrainer_greenskin_open_book_str}I'z here fer sum ju-ju spooks." />
+	<string id="tor_spelltrainer_greenskin_close_book_str" text="{=tor_spelltrainer_greenskin_close_book_str}Eat dis shroom an' dance wiv me." />
+	<string id="tor_spelltrainer_greenskin_enchantments_str" text="{=tor_spelltrainer_greenskin_enchantments_str}I wanna learn how ta make some proppa gear wif da runes." />
+	<string id="tor_spelltrainer_greenskin_player_goodbye_str" text="{=tor_spelltrainer_greenskin_player_goodbye_str}I'z outta 'ere." />
+	<string id="tor_spelltrainer_greenskin_goodbye_str" text="{=tor_spelltrainer_greenskin_goodbye_str}Get outta 'ere! You'ze dispersin' my smoke." />
+
+	<!--Vanilla perk changes-->
+    <string id="GJRcFc0V" text="{=GJRcFc0V}A variable chance to recover a mount from lost cavalry after battles." />
```
SlyDevil: I'm unsure how the gametextmanager will store/fetch this in relation to the pre-existing string defined in DefaultsPerks.

Does this modified string show correctly in english and another arbitrary language (should also show the english string in that language)?

If yes, fine by me for the moment.

SlyDevil (approved):
Final comments for minor changes.
I'm curious what the drops will look like from raids, but there's no immediate problem.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
SlyDevil: I see what you mentioned with the blueprints getting added previously; I missed the foreach loop that handles it.

CSharpSourceCode/HarmonyPatches/VeterinarianPatch.cs
```diff
@@ -53,4 +57,52 @@ private static float GetMountRecoveryChance(ItemObject mountItem)
 
         return NORMAL_MOUNT_RECOVERY_CHANCE;
     }
+
+    [HarmonyPatch]
+    internal static class VeterinarianPerkDescriptionPatch
+    {
+        private const string VETERINARIAN_PERK_ID = "MedicineVeterinarian";
+        private const string SECONDARY_DESCRIPTION_REPLACEMENT =
+            "{=tor_veterinarian_description}A variable chance to recover a mount from lost cavalry after battles.";
```
SlyDevil: str_tor_veterinarian_description

Edit : I'll handle this and adding it to strings xml after merge.

SlyDevil (approved):
Last comment for the localization id.

Can merge after.

CSharpSourceCode/Quests/Careers/OrcBossQuest2.cs
```diff
@@ -15,13 +15,15 @@ namespace TOR_Core.Quests.Careers
...
-        private const int RequiredBrawlsWon = 100;
+        private const int RequiredWeaponSkillLevels = 250;
+        private const int RequiredHighRenownBattles = 3;
+        private const int RequiredSingleBattleRenown = 200;
+        private const int RequiredArenaFights = 15;
+        private const int RequiredBrawlsWon = 20;
         private const int RequiredCitiesCaptured = 5;
         private const int RequiredLordDuels = 15;
-        private const int RequiredTeefTransferred = 500000;
+        private const int RequiredTeefTransferred = 1000000;
```
Z3rca: i would review this with gamedesign

CSharpSourceCode/Quests/Careers/OrcBossQuest1.cs
```diff
@@ -163,9 +163,29 @@ private void OnSkillIncreased(Hero hero, SkillObject skill, int skillValueBefore
 
         private void OnMapEventEnded(MapEvent mapEvent)
         {
-            _currentBattlesWon++;
-            _taskBattlesWon.UpdateCurrentProgress(_currentBattlesWon);
-            UpdateQuest();
+            if (_currentBattlesWon == 1)
```
Z3rca: there is a bug i see it but what does this do?

CSharpSourceCode/Quests/Careers/OrcBossQuest1.cs
```diff
@@ -14,11 +14,11 @@ namespace TOR_Core.Quests.Careers
     public class OrcBossQuest1 : QuestBase
     {
         // Quest requirements constants
-        private const int RequiredWeaponSkillLevels = 150;
-        private const int RequiredBattlesWon = 50;
-        private const int RequiredArenaFights = 20;
-        private const int RequiredTeefTransferred = 150000;
-        private const int RequiredBrawlsWon = 50;
+        private const int RequiredWeaponSkillLevels = 175;
```
Z3rca: again gamedesign debate.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
Z3rca, line 1: i would like to have a bit more description here, what is your intention, for this? this seems to work but has this been discussed with game design?

Z3rca (changes requested):
overall I dont see much issues, rather than it requires more detailed explanation WHY those changes are made. I dont see a description in the pr, nor some comments of yours inside the code why balance changes were made. please act on these , maybe consult piero. I would require you to check this with designers before it gets approved.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
emrebahat: excluded artifacts/legendary bows, it was a request
blueprints are now explicitly added into the *items*, so other checks won't work on them, i believe this was the reason why their drop rate felt incredibly low
after you get any blueprints, he number of equipment picks is reduced to keep the total roughly constant. this could (and i think should) be adjusted furthermore.
older culture check was bugged, also didn't exclude crafted items/vanilla items for overriden cultures.
old code had weapons that could still get more traits even if already traited, shields weren't included and armor couldn't if traited
tier checks has been added so you are less likely to end up with some crap of your culture
iirc my pr log also mentioned them. if you want me to add comments to a specific part of the code or any of these changes to be discussed with game design lmk.

Z3rca (approved):

PR #5, opened 2025-12-22
Removal of Damage patch and refactoring damage logic

Z3rca:
This branch removes the Harmony-based damage patch system entirely. All damage modifications—melee, ranged, and spell—are now handled through the game's model system following vanilla Bannerlord patterns.

  Melee and ranged damage modifications have been moved to TORAgentApplyDamageModel, extending SandboxAgentApplyDamageModel. This follows the vanilla model pattern where game logic is encapsulated in replaceable model classes. Unfortunately, melee and ranged damage no longer use custom display—they rely on vanilla UI. On the bright side, this eliminates the previous offset between our calculated values and what the game displayed, and results in less clutter and confusion in the combat log.

  Old Way
  ```mermaid
  flowchart TD
      A[All Damage Types] --> B[Mission.GetAttackCollisionResults]
      B --> C[AgentApplyDamageModel.CalculateDamage]
      C --> D[Harmony DamagePatch]
      D --> E[Mission.RegisterBlow]
      E --> F[Agent.HandleBlow]
      F --> G[OnScoreHit CombatXpModel]
  ```
  New Way: Melee & Ranged
  ```mermaid
  flowchart TD
      A[Melee and Ranged] --> B[Mission.GetAttackCollisionResults]
      B --> C[TORAgentApplyDamageModel.CalculateDamage]
      C --> D[blow.DamageCalculated = true]
      D --> E[Mission.RegisterBlow]
      E --> F[Agent.HandleBlow]
      F --> G[Vanilla Display and XP]
  ```
  New Way: Spell Damage & Healing
  ```mermaid
  flowchart TD
      A[Spell Cast] --> B[TriggeredEffect.Trigger]
      B --> C{DamageAmount}
      C -->|Positive| D[CalculateAbilityDamage]
      C -->|Negative| E[CalculateAbilityHealing]
      D --> F[Agent.ApplyDamage]
      E --> G[Agent.Heal]
      F --> H[RegisterBlow]
      H --> I[HandleBlow]
      I --> J[SpellCastSession]
      G --> J
      J --> K[FinalizeSession]
      K --> L[Spellcraft Faith XP]
  ```
  ```

  Spell damage and healing calculations are now handled in TORAbilityModel, with virtual hooks (ApplyDamageModifiers, ApplyHealingModifiers) for extensibility. The calculation path is split early: positive DamageAmount routes to CalculateAbilityDamage, negative values (instant healing like Storm of Renewal) route to CalculateAbilityHealing. Spell damage uses a flag to skip vanilla damage calculations entirely, cleanly separating it from melee/ranged pathways. This also fixes XP attribution—Athletics and Riding XP are no longer incorrectly granted as a side effect of dealing spell damage.

  A new SpellCastSession system aggregates all effects from a single spell cast—damage, healing, kills, and status effects—enabling deferred XP granting and end-of-spell result display. Sessions persist until status effects (DOTs/HOTs) expire, ensuring kills from delayed effects are properly attributed. This also allows for more engaging Spellcraft/Faith XP gain: XP now scales with how many units were affected, how much damage or healing was dealt, and how many status effects were applied—rewarding effective spellcasting.

<img width="657" height="340" alt="image" src="https://github.com/user-attachments/assets/3a5d1d8d-1765-46f6-86b7-ed3aaec6e48e" />

_after the SpellCast session expired (after dots were applied), the total damage dealed is calculated and displayed._



<img width="738" height="111" alt="image" src="https://github.com/user-attachments/assets/7d254cfb-a376-4c12-b650-46665ea6a102" />

_direct damages are displayed and also killed units_

<img width="601" height="337" alt="image" src="https://github.com/user-attachments/assets/dc7007eb-34e2-4aa4-8b01-753241ca0a54" />

_healing is also displayed._
 

---
  1. Damage Patch Removal

  The Harmony-based DamagePatch has been removed entirely. Previously, this patch intercepted damage calculations to apply TOR-specific modifiers, but this approach had several issues:
  - Offset between calculated and displayed damage values
  - Athletics/Riding XP incorrectly granted for spell damage
  - Difficult to maintain and extend
  - Cluttered combat log with redundant messages

  2. Melee & Ranged Damage: TORAgentApplyDamageModel

  All melee and ranged damage modifications now live in TORAgentApplyDamageModel, which extends SandboxAgentApplyDamageModel.

  Key methods:
  - CalculateDamage() - Applies TOR-specific damage amplifications and resistances
  - CalculateWardSaveFactor() - Calculates ward save chance based on resistance percentages
  - CreateAgentPropertyContainer() - Builds property containers with damage/resistance percentages

  Property System:
  - DamagePercentages - Base damage amplification per damage type
  - AdditionalDamagePercentages - Bonus damage from perks/careers
  - ResistancePercentages - Damage reduction per damage type

  3. Spell Damage & Healing: TORAbilityModel

  Public calculation methods:
  - CalculateAbilityDamage(attacker, victim, baseDamage, damageType, abilityTemplate) - For damage spells
  - CalculateAbilityHealing(caster, target, baseHealing, abilityTemplate) - For healing spells

  Virtual hooks for extensibility:
  - ApplyDamageModifiers(...) - Applies perk effects, skill effectiveness, career bonuses
  - ApplyHealingModifiers(...) - Applies skill effectiveness to healing

  Routing logic in TriggeredEffect.Trigger():
  if (DamageAmount > 0)  → TORMissionHelper.DamageAgents() → CalculateAbilityDamage()
  if (DamageAmount < 0)  → TORMissionHelper.HealAgents()  → CalculateAbilityHealing()

  Spell damage flag:
  Spell damage calls Agent.ApplyDamage() with originatesFromAbility: true, which skips vanilla damage calculations and prevents incorrect XP attribution.

  4. SpellCastSession System

  SpellCastSession.cs - Tracks per-cast aggregates:

  | Property             | Description                                            |
  |----------------------|--------------------------------------------------------|
  | CastID               | Unique identifier for this cast                        |
  | Caster / CasterHero  | The casting agent and hero reference                   |
  | AbilityTemplate      | The spell/prayer being cast                            |
  | TotalDamageDealt     | Sum of all damage dealt                                |
  | TotalHealingDone     | Sum of all healing done                                |
  | AgentsDamagedCount   | Unique agents damaged                                  |
  | AgentsHealedCount    | Unique agents healed                                   |
  | AgentsKilledCount    | Unique agents killed                                   |
  | StatusEffectsApplied | Count of status effects applied                        |
  | EarliestCollectTime  | When session can be finalized (after DOTs/HOTs expire) |

  Booking methods:
  - BookDamage(victim, damageDealt, damageAbsorbed, damageType)
  - BookHealing(target, healingDone)
  - BookKill(victim)
  - BookStatusEffect(target)
  - ExtendCollectTime(duration) - Delays collection for DOTs/HOTs

  Session lifecycle in AbilityManagerMissionLogic:
  1. StartSpellCastSession(caster, abilityTemplate) - Creates session, returns castId
  2. BookSpellDamage/Healing/Kill/StatusEffect() - Called during spell execution
  3. TryCollectSession() - Called on tick, collects if IsReadyToCollect
  4. FinalizeSession() - Grants XP, shows display, removes session
  5. FinalizeAllPendingSessions() - Called on battle end via OnMissionResultReady

  Why store CasterHero?
  The Agent reference may become invalid after battle ends, but we still need to grant XP. The Hero reference is captured at session creation.

  5. XP Calculation

  CalculateSpellSessionXp(SpellCastSession session):
  int xp = 0;
  xp += session.TotalDamageDealt / 5;
  xp += session.TotalHealingDone / 5;
  xp += session.AgentsAffectedByStatusEffectsCount * 10;
  return xp;

  XP is granted to the relevant skill (TORSkills.Spellcraft or TORSkills.Faith) based on AbilityTemplate.AbilityType.

  6. Localization

  Spell result display uses TORTextHelper with TextObject variable system for proper localization:
  - String keys defined in tor_strings.xml
  - Variables like {DAMAGE}, {HEALING}, {KILLS} substituted at runtime


CSharpSourceCode/BattleMechanics/StatusEffect/StatusEffectComponent.cs
```diff
@@ -136,18 +141,25 @@ public void OnElapsed(float dt)
...
+                    var damageValue = (int)_effectAggregate.DamageOverTime;
                     var effect = _currentEffects.Keys.FirstOrDefault(x => x.Template.Type == StatusEffectTemplate.EffectType.DamageOverTime && x.ApplierAgent == Agent.Main) ??
                                   _currentEffects.Keys.FirstOrDefault(x => x.Template.Type == StatusEffectTemplate.EffectType.DamageOverTime);
 
                     var applier = effect?.ApplierAgent;
 
                     if (Campaign.Current != null && applier != null && (applier.IsMainAgent || applier.BelongsToMainParty()))
                     {
-                        CareerHelper.ApplyCareerAbilityCharge((int)value, ChargeType.Healed, AttackTypeMask.Spell, applier);
+                        CareerHelper.ApplyCareerAbilityCharge(damageValue, ChargeType.Healed, AttackTypeMask.Spell, applier);
```
SlyDevil: Does the ChargeType matter here?

DamageOverTime is positive which would suggest a type unrelated to healing.

Z3rca: this is for the Charge Suppliers... for example damsels, yes it does matter, because the code runs in both cases, heal or damage.

Z3rca: now i see. it should be damage done. yes thats wrong here, mattered and  have caused many issues :D

Z3rca: fixed on development and here

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -69,6 +77,67 @@ public int GetSkillXpForAbilityDamage(AbilityTemplate ability, int damageAmount)
...
+        /// </summary>
+        public int CalculateSpellSessionXp(SpellCastSession session)
+        {
+            if (session == null)
+                return 0;
+
+            int xp = 0;
+
+            // XP for damage dealt (damage / 5)
+            if (session.TotalDamageDealt > 0)
```
SlyDevil, line 92: We may want a balancing factor in here for single/aoe spells.

Fireball is going to blast everything in a 6m(?) radius racking up big damage, and getting so many status applications.

Spells that target only allies may also be worth giving a bonus to as they're unable to benefit from the damage xp, and even if they grant healing, healing tends to be lower numbers than damage.

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -595,5 +664,266 @@ public bool IsValidLoreForCharacter(Hero hero, LoreObject loreObject)
...
+            // Apply amplifications and resistances
+            damageAmplifications[damageTypeIndex] += additionalDamagePercentages[damageTypeIndex];
+            damageAmplifications[damageTypeIndex] -= resistancePercentages[damageTypeIndex];
+            resultDamage *= (1 + damageAmplifications[damageTypeIndex]);
+
+            // Apply ward save
+            resultDamage *= wardSaveFactor;
+
+            // Clamp to prevent negative damage on enemies
+            if (!friendlyFire && resultDamage < 0)
```
SlyDevil: Max(0, negative number) clamps all resultDamages within [0, MaxDamage] anyways; no need to use 2 clamps.

The friendly fire check is unneeded.

If this feeds into xp anywhere, I'd say let this handle the negative damage, then have the end return instead be :
`Math.Min(agent.RemainingHealth, resultDamage)`
to avoid granting xp for overkill damage (unless this is handled somewhere I haven't seen yet.

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -595,5 +664,266 @@ public bool IsValidLoreForCharacter(Hero hero, LoreObject loreObject)
...
+                        if (MobileParty.MainParty.LeaderHero.HasAnyCareer())
+                        {
+                            if (Hero.MainHero.HasCareerChoice("AncientScrollsPassive4"))
+                            {
+                                damageAmplifications[damageTypeIndex] += 0.2f;
+                            }
+                        }
+                    }
+
+                    if (attacker.HasAttribute("Arcane_Dmg"))
```
SlyDevil, line 887: Is this a spell damage attribute for npcs?

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -595,5 +664,266 @@ public bool IsValidLoreForCharacter(Hero hero, LoreObject loreObject)
...
+            if (Game.Current.GameType is not Campaign)
+                return healing;
+
+            if (abilityTemplate != null && caster.IsHero)
+            {
+                var hero = caster.GetHero();
+                if (hero != null)
+                {
+                    // Skill effectiveness for healing
+                    healing *= GetSkillEffectivenessForAbilityDamage(hero.CharacterObject, abilityTemplate);
```
SlyDevil, line 914: Perks like Overcharged Caster (? the one that gives +damage + wom cost), should they also apply to healing effects?

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -595,5 +664,266 @@ public bool IsValidLoreForCharacter(Hero hero, LoreObject loreObject)
...
+
+            int damageTypeIndex = (int)damageType;
+
+            if (abilityTemplate != null && attacker.IsHero)
+            {
+                var hero = attacker.GetHero();
+                if (hero != null)
+                {
+                    // Perk effects multiplier
+                    damage *= GetPerkEffectsOnAbilityDamage(hero.CharacterObject, victim, abilityTemplate);
```
SlyDevil, line 849: Career perks like greylord's that give +X% effectiveness (damage boost if no aoes equipped) under certain conditions, are those added here, or elsewhere when calculating damage?

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -595,5 +664,266 @@ public bool IsValidLoreForCharacter(Hero hero, LoreObject loreObject)
...
+
+                if (baseDamage <= 0) continue;
+
+                // Calculate final damage with all modifiers
+                int finalDamage = CalculateAbilityDamage(caster, agent, baseDamage, damageType, abilityTemplate);
+
+                if (finalDamage > 0)
+                {
+                    // Apply the damage
+                    agent.ApplyDamage(finalDamage, impactPosition, caster, doBlow: true, hasShockWave: hasShockWave, originatesFromAbility: abilityTemplate != null);
```
SlyDevil, line 964: The equivalent healing method adds career charge at the end, is the charging from damage dealt handled elsewhere?

I vaguely remembering seeing it, but unsure atm.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -271,6 +271,126 @@ public override bool DecideMountRearedByBlow(Agent attackerAgent, Agent victimAg
...
+            var attackerAgent = attackInformation.IsAttackerAgentMount ? attackInformation.AttackerAgent?.RiderAgent : attackInformation.AttackerAgent;
+            var victimAgent = attackInformation.IsVictimAgentMount ? attackInformation.VictimAgent?.RiderAgent : attackInformation.VictimAgent;
+
+            if (attackerAgent == null || victimAgent == null)
+                return vanillaDamage;
+
+            if (!attackerAgent.IsHuman || !victimAgent.IsHuman)
+                return vanillaDamage;
+
+            if (attackerAgent == victimAgent)
```
SlyDevil, line 382: Interesting, the player nuking themself won't benefit from their resistances or penalize them for their amplifications.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -271,6 +271,126 @@ public override bool DecideMountRearedByBlow(Agent attackerAgent, Agent victimAg
...
+                return true;
+
+            // TOR-specific shrug off logic
+
+            // Unstoppable attribute (e.g., large monsters, bosses)
+            if (victimAgent.HasAttribute("Unstoppable"))
+                return true;
+
+            // Agent's own damage threshold check
+            if (victimAgent.IsDamageShruggedOff(blow.InflictedDamage))
```
SlyDevil, line 462: IsDamageShruggedOff is where career passives like sigmar's that ignores sufficiently low damage blows are handled, iirc?

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
```diff
@@ -804,6 +817,207 @@ public override void OnEarlyAgentRemoved(Agent affectedAgent, Agent affectorAgen
...
+                if (model != null)
+                {
+                    var xpAmount = model.CalculateSpellSessionXp(session);
+                    if (xpAmount > 0)
+                    {
+                        var skill = model.GetRelevantSkillForAbility(session.AbilityTemplate);
+                        session.CasterHero.AddSkillXp(skill, xpAmount);
+
+                        // DarkVisionPassive3 - also grants Roguery XP
+                        if (session.CasterHero.HasAnyCareer() && session.CasterHero.HasCareerChoice("DarkVisionPassive3"))
```
SlyDevil: I thought I saw xp from this career perk being handled elsewhere as well, but I can't find it and I'm unsure if that's because I'm confusing 2 classes or if I'm bad a searching atm.

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
```diff
@@ -804,6 +817,207 @@ public override void OnEarlyAgentRemoved(Agent affectedAgent, Agent affectorAgen
...
+                }
+            }
+
+            // Grant XP for all hero casters (player and companions)
+            if (Game.Current.GameType is Campaign && session.CasterHero != null)
+            {
+                var model = Campaign.Current.Models.GetAbilityModel();
+                if (model != null)
+                {
+                    var xpAmount = model.CalculateSpellSessionXp(session);
```
SlyDevil: Was friendly fire damage getting filtered out elsewhere, or will nuking your own troops be simultaneously a source of spellcraft _and_ medicine xp?

SlyDevil:
I've commented on potential issues I saw.

I like the move away from having to handle stuff in a harmony patch.
I wonder what it shows in game if you turn on the detailed damage record in settings and hit someone with a spell.

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: Its from the Magister a Damage perk. there was a reason i put it there this way. i just dont remember why... long time ago

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
Z3rca: goo catch. i dont think i accounted for this.

CSharpSourceCode/Models/TORAbilityModel.cs
SlyDevil: The status effect `arcane_conduit_dmg_buff` grants it as a temp attribute.

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: sure we can do that. but thats also because we are now able to.

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: i guess they should be there yes. if not we should move it there

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
Z3rca: oh thats an residual of fall damage of the regular damage model. i assume this wont be applicable anymore if those are now seperated.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
Z3rca: yes, it is maybe not applicable here. however we need to consider status effects, that would let them "shrug" everytime you apply the damage. i have to think about this more

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
Z3rca: So there should be two positions: One cast position, and one at the end of the session collection. The onCast xp is only for the XP you gain from casting directly , or prayers that suppose to only give you xp while casting. 

since we can track better healing and overall damage application etc, i think we should keep it as it is in this pull request that the dealed damage / healing is booked in the end also for faith.

CSharpSourceCode/AbilitySystem/SpellCasting/SpellCastSession.cs
```diff
@@ -57,15 +57,22 @@ public SpellCastSession(int castId, Agent caster, AbilityTemplate abilityTemplat
...
         /// <summary>
         /// Books damage dealt to an agent in this session.
+        /// Friendly fire damage is tracked but not counted towards XP.
         /// </summary>
         public void BookDamage(Agent victim, int damageDealt, int damageAbsorbed, DamageType damageType)
         {
-            if (victim != null)
+            if (victim == null) return;
+
+            bool isFriendlyFire = Caster != null && victim.Team == Caster.Team;
```
SlyDevil, line 72: Can an agent be teamless?

A mount maybe?

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: you get that attribute from there yes. to be fair ,  i am not sure anymore why it is there... i think i change this now.

Z3rca: the code is a bit legacy , I just dont think we have a better way, since any magic damage should be enhanced. I could rework the entire code, or just leave here an ugly check. i rather stick to the latter.

CSharpSourceCode/AbilitySystem/SpellCasting/SpellCastSession.cs
Z3rca: you cant apply spellcasting to mounts so far.

CSharpSourceCode/Models/TORStrikeMagnitudeModel.cs
```diff
@@ -72,11 +70,11 @@ public override float CalculateAdjustedArmorForBlow(in AttackInformation attackI
                         }
                     }
 
-                    if (agent.WieldedWeapon.CurrentUsageItem != null && agent.WieldedWeapon.CurrentUsageItem.GetItemUsageIndex() == weaponComponent.GetItemUsageIndex())
+                    if (attackerAgent.WieldedWeapon.CurrentUsageItem != null && attackerAgent.WieldedWeapon.CurrentUsageItem.GetItemUsageIndex() == weaponComponent.GetItemUsageIndex())
```
SlyDevil, line 73: When exhausting ammo for a thrown weapon, does the game automatically swap to the next weapon/change the wielded weapon to empty?
Ie. if you throw your last enchanted grenade/axe/whatever, would its traits not apply?

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
Z3rca: yes checked its good. the career perks with shrugged off are applied correctly

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
Z3rca: This only affects weapon-based damage (melee/ranged), not spell damage. Self-weapon-damage scenarios are rare (grenades, fall damage) and returning vanilla damage is imo acceptable then.

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
Z3rca: its actually 7 positions, but i checked all of them. the application of the charge should be fine for all positions

CSharpSourceCode/Models/TORStrikeMagnitudeModel.cs
Z3rca: Fixed. Added a switch on missileWeapon.CurrentUsageItem.WeaponClass to properly source traits from the missile weapon itself for thrown weapons, while ranged weapons (bows/guns) still check the wielded weapon for traits.

Z3rca: fixed

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: i added a comment , this doesnt need to be added here or would be applied twice

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
Z3rca: we added friendly fire now to the booking for testing

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: this should be handled

career perks using PassiveEffectType.SpellEffectiveness (like Necrarch's +15% with light armor, Imperial Magister's +15% with offhand staff) are handled in GetPerkEffectsOnAbilityDamage() at line 397 via CareerHelper.ApplyBasicCareerPassives(). 

UnrestrictedMagicPassive4 is also handled in GetSkillEffectivenessForAbilityDamage

CSharpSourceCode/Models/TORAbilityModel.cs
Z3rca: i added 5 x xp bonus for single target spells

PR #6, opened 2026-01-03
Performance 2

alemreM:


CSharpSourceCode/BattleMechanics/TriggeredEffect/TriggeredEffect.cs
```diff
@@ -174,6 +174,10 @@ private void PlaySound(Vec3 position)
             if (_template != null && _template.SoundEffectId != "none")
             {
                 _soundIndex = SoundEvent.GetEventIdFromString(_template.SoundEffectId);
+                if (_soundIndex == -1)
```
Z3rca: i am not sure if there was a thought in this. what if a sound is tried to be retrieved many many times, but it doesnt exist. by having a crash you can atleast figure out what was wrong. so atleast log here the wrong soundEffectId, since it shouldn't be -1

CSharpSourceCode/BattleMechanics/TriggeredEffect/TriggeredEffect.cs
```diff
@@ -102,7 +102,7 @@ public void Trigger(Vec3 position, Vec3 normal, Agent triggererAgent, AbilityTem
                 {
                     if (triggererAgent.Character is CharacterObject triggererCharacter && triggererCharacter.GetPerkValue(TORPerks.Spellcraft.ArcaneLink) && effect.IsBuffEffect)
                     {
-                        if (!targets.Contains(triggererAgent)) targets.Append(triggererAgent);
+                        if (!targets.Contains(triggererAgent)) targets.Add(triggererAgent);
```
Z3rca, line 105: good catch. Append wouldnt affect the targets, while add allows to add them. I searched the codebase for other issues like this but couldnt find them.

CSharpSourceCode/BattleMechanics/TriggeredEffect/TriggeredEffect.cs
```diff
@@ -185,7 +189,15 @@ private void TriggerScript(Vec3 position, Agent triggerer, IEnumerable<Agent> tr
...
-                    var obj = Activator.CreateInstance(Type.GetType(_template.ScriptNameToTrigger));
+                    var scriptType = Type.GetType(_template.ScriptNameToTrigger, throwOnError: false);
+                    if (scriptType == null)
+                    {
+                        TORCommon.Log("Tried to spawn TriggeredScript: " + _template.ScriptNameToTrigger + ", but type could not be resolved.", NLog.LogLevel.Error);
+                        return;
+                    }
+
+
+                    var obj = Activator.CreateInstance(scriptType);
```
Z3rca, line 196: good thinking.

CSharpSourceCode/CampaignMechanics/Companions/TORCompanionsCampaignBehavior.cs
```diff
@@ -214,18 +216,32 @@ private void CanHeroDie(Hero dyingHero, KillCharacterAction.KillCharacterActionD
...
+                }
+
+                if (!clanmember.GetPerkValue(TORPerks.Spellcraft.StoryTeller))
+                {
+                    continue;
+                }
+
+                foreach (var companion in clanmember.CompanionsInParty)
+                {
+                    var randomSkillIndex = MBRandom.RandomInt(skillCount);
```
Z3rca: we need to catch here suitable skills. a vampire companion should not learn about faith or non magical characters shouldnt learn about spellcraft

CSharpSourceCode/CampaignMechanics/CustomResources/CustomResourceManager.cs
```diff
@@ -833,7 +833,7 @@ public static void OnTroopTransferred(PartyVM partyVm, PartyScreenLogic.PartyRos
 
             AddResourceChanges(Hero.MainHero.GetCultureSpecificCustomResource(), sign * (int)explainedNumber.ResultNumber);
 
-            partyVm.GetExtensionInstance().RefreshValues();
+            partyVm.GetExtensionInstance()?.RefreshValues();
```
Z3rca, line 836: how should this be null? it should rather crash if for what ever reasons that is null.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -429,7 +429,7 @@ private void OnMissionEnded(IMission obj)
                     var text = new MBStringBuilder();
                     text.Initialize();
 
-                    var hintInfo = item.GetTorSpecificData();
+                    var hintInfo = item.GetTorSpecificDataReadOnly();
```
Z3rca, line 432: it makes sense to make the TorSpecificDataReadOnly. , did you investigate the cases where you actually need a manipulatable object( that gets put back to the extended item infos? I dont see the purpose of having non- Read-only elements... Maybe the Enchantment system needs it? I would investigate this and bundle the cases. the "default" should be read only.

CSharpSourceCode/Extensions/ExtendedInfoSystem/ExtendedInfoManager.cs
```diff
@@ -235,18 +235,24 @@ private void OnNewGameCreatedPartialFollowUpEnd(CampaignGameStarter campaignGame
         private void HourlyTick() //I wonder if this could be done on the quarter daily tick instead and reduce the load significantly? The quarter tick is party-based, it could be performed on the main party's quarter tick, ie QuarterDailyTick above.
         {
             //fill winds of magic
-            foreach (var entry in _heroInfos)
+            foreach (var hero in Hero.AllAliveHeroes)
```
Z3rca, line 238: while this is true, i dont see here a noteworthy improvement. Slys former comment does point something out that could be beneficial. we are currently checking way too often every Hero. we could limit the important checks to player clan only, while all other have a quarter daily tick that batches the gains. this should make these calls a little bit less extensive. the Hero.AllAlliveHeroes dont increase significantly the performance i feel.

CSharpSourceCode/Extensions/ItemObjectExtensions.cs
```diff
@@ -71,19 +82,18 @@ public static bool IsInventoryUsable(this ItemObject item)
 
         public static List<ItemTrait> GetTraits(this ItemObject item)
```
Z3rca, line 106: I wonder if a bool return value, with a out parameter of the traits wouldn't be the better way to go. This could enhance the performance post retrieving traits since the code after retrieval  could be mostly avoided.

CSharpSourceCode/Extensions/ItemObjectExtensions.cs
```diff
@@ -124,11 +138,10 @@ public static ExtendedItemObjectProperties GetTorSpecificData(this ItemObject it
 
         public static bool HasAnyTrait(this ItemObject item)
```
Z3rca, line 162: this code could then be avoided (see above)

CSharpSourceCode/Extensions/UI/SPItemVMExtension.cs
```diff
@@ -18,16 +18,19 @@ namespace TOR_Core.Extensions.UI
...
         public SPItemVMExtension(ViewModel vm) : base(vm)
         {
-            _useHint = new HintViewModel(TORTextHelper.GetTextObject("tor_item_hint_use", "Use Item"));
+            _sharedUseHint ??= new HintViewModel(TORTextHelper.GetTextObject("tor_item_hint_use", "Use Item"));
+            _useHint = _sharedUseHint;
         }
 
         public override void RefreshValues()
         {
+            base.RefreshValues();
```
Z3rca: are you use this is required? base ViewModelExtension doesnt do anything as far as a I am aware?

CSharpSourceCode/Extensions/ViewModelExtensions.cs
```diff
@@ -10,9 +10,35 @@ namespace TOR_Core.Extensions
 {
     public static class ViewModelExtensions
     {
-        public static bool HasExtensionType(this ViewModel model) => ViewModelExtensionManager.Instance.HasViewModelExtensionType(model);
-        public static bool HasExtensionInstance(this ViewModel model) => ViewModelExtensionManager.Instance.HasViewModelExtensionInstance(model);
-        public static IViewModelExtension GetExtensionInstance(this ViewModel model) => ViewModelExtensionManager.Instance.GetExtensionInstance(model);
-        public static Type GetExtensionType(this ViewModel model) => ViewModelExtensionManager.Instance.GetExtensionType(model);
+        public static bool HasExtensionType(this ViewModel model)
```
Z3rca, line 13: I dont see the benefit of these changes. if it crashs then it crashs. you either log the wrong or non working viewmodel, otherwise you just bandaid errors here without an option to analyse the underlying issue.

CSharpSourceCode/HarmonyPatches/InventoryPatches.cs
Z3rca, line 1: Might be style thing. what is the benefit of outsourcing the intern methods to some static fields? In the best case, we should avoid Patches as much as we can, while adding such fields sounds a bit like decorating your home. it makes also later deactivations a bit harder, sicne you might need to shotgun surgeon your way through the patch. this is not a lot difference, but if we are not benefiting from a performance boost or similar, i dont see here a massive benefit in doing this.

CSharpSourceCode/HarmonyPatches/InventoryPatches.cs
```diff
@@ -141,6 +168,8 @@ private static Color GetColorForDamageType(DamageType type)
         private static void EquipItemsFromClipboard(List<Tuple<string, int>> itemList, InventoryLogic logic, SPInventoryVM inventoryVM)
         {
             List<TransferCommand> commands = new List<TransferCommand>();
+            var playerItems = logic.GetElementsInRoster(InventoryLogic.InventorySide.PlayerInventory);
+            var otherItems = logic.GetElementsInRoster(InventoryLogic.InventorySide.OtherInventory);
```
Z3rca, line 174: this might improve the performance since we are not reretrieving the items all the time. good catch

CSharpSourceCode/HarmonyPatches/PartyScreenPerformancePatches.cs
```diff
@@ -0,0 +1,142 @@
...
+namespace TOR_Core.HarmonyPatches
+{
+
+    [HarmonyPatch(typeof(PartyVM), "TransferAllTroops")]
+    internal static class PartyVM_TransferAllTroops_PerformancePatch
+    {
+        private static readonly Action<PartyVM> RefreshPartyInformation =
+            AccessTools.MethodDelegate<Action<PartyVM>>(AccessTools.Method(typeof(PartyVM), "RefreshPartyInformation"));
+
+        private static bool Prefix(PartyVM __instance, PartyScreenLogic.PartyCommand command)
```
Z3rca, line 22: Explain for every patch that you add in detail what this does and for what purpose this was done. i dont want any patches in the code anymore that are not documented. We should improve ourselves in this regard

CSharpSourceCode/HarmonyPatches/SettlementPatches.cs
```diff
@@ -128,10 +128,37 @@ public static void AddExtrasToSettlementInfo(PropertyBasedTooltipVM __instance,
 
         [HarmonyPrefix]
         [HarmonyPatch(typeof(SettlementNameplatesVM), "AddNameplate")]
-        public static bool AddCustomNamePlateVM(SettlementNameplatesVM __instance, SettlementNameplateVM nameplate, 
+        public static bool AddCustomNamePlateVM(SettlementNameplatesVM __instance, SettlementNameplateVM nameplate,
```
Z3rca, line 136: also add for patches you now dived into again, the description. this would help us to document the reasoning behind it.

CSharpSourceCode/HarmonyPatches/ViewModelPatches.cs
Z3rca, line 1: i need more context for this patch. I understood the former way, but you filter now between Properties and Methods , i am not quite sure what the benefit is here. why this was done.

CSharpSourceCode/Items/TorItemMenuVM.cs
```diff
@@ -217,8 +217,8 @@ private void AddTooltipForRaceLock(SPItemVM item, ItemVM comparedItem)
             var equipmentElement = item.ItemRosterElement.EquipmentElement;
             if (!equipmentElement.Item.HasArmorComponent || equipmentElement.Item.ItemType == ItemObject.ItemTypeEnum.HorseHarness) return;
             var comparedEquipmentElement = comparedItem == null ? EquipmentElement.Invalid : comparedItem.ItemRosterElement.EquipmentElement;
-            var info = equipmentElement.Item.GetTorSpecificData();
-            CreateColoredProperty(TargetItemProperties, GameTexts.FindText("str_inventory_race_lock").ToString(), info.RaceLock.ToString(), GetColorFromBool(ExtendedItemObjectManager.CanCharacterUseItemBasedOnRace(equipmentElement.Item, CurrentCharacter)), 0, null, TooltipProperty.TooltipPropertyFlags.None);
+            var raceLock = equipmentElement.Item.GetTorSpecificDataReadOnly()?.RaceLock ?? "human";
```
Z3rca: use the TorTextHelper to allow for displaying any racename. A textObject with variant should be good for this.

Z3rca (changes requested):
Overall some interesting and good improvements, yet i think there should be a bit more documentation for the patches and please react to the comments i made.

CSharpSourceCode/CampaignMechanics/CustomResources/CustomResourceManager.cs
emrebahat: > how should this be null? it should rather crash if for what ever reasons that is null.

PartyVM can be null in some transfer flows/teardwn timing, especially in longer sessions. so id rather keep this as a best effort ui refresh instead of crashing the party screen. i can add a warning log if that's what you prefer

CSharpSourceCode/Extensions/ItemObjectExtensions.cs
emrebahat: > I wonder if a bool return value, with a out parameter of the traits wouldn't be the better way to go. This could enhance the performance post retrieving traits since the code after retrieval could be mostly avoided.

hmm. i will also cache trait lookups by string id to avoid scanning ItemTrait.All repeatedly

CSharpSourceCode/Extensions/UI/SPItemVMExtension.cs
emrebahat: > are you use this is required? base ViewModelExtension doesnt do anything as far as a I am aware?

my bad

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
emrebahat: > it makes sense to make the TorSpecificDataReadOnly. , did you investigate the cases where you actually need a manipulatable object( that gets put back to the extended item infos? I dont see the purpose of having non- Read-only elements... Maybe the Enchantment system needs it? I would investigate this and bundle the cases. the "default" should be read only.
theres currently no path depending on a mutable object thru this api

CSharpSourceCode/Extensions/ExtendedInfoSystem/ExtendedInfoManager.cs
emrebahat: the point isnt just switching to AllAliveHeroes, its removing per entry Hero.FindFirst(..) call inside the loop (that was effectively ON^2 on hourly ticks as hero count goes up). now its a single pass with ictionary lookup.

CSharpSourceCode/Extensions/ViewModelExtensions.cs
emrebahat: these werent meant to hide bugs, just convenience''try-get'' extension helpers that can legit be called during gauntlet screen teardown/refresh when the vm reference is a temp null. returning false/null here prevents ui reload crashes (again for longer sessions and when cheat menü is enabled)
i removed manager != null checks so the only guard left is model == null, lmk what do u think.

CSharpSourceCode/HarmonyPatches/InventoryPatches.cs
emrebahat: this patch is already running in a hot ui path (inv OnFrameTick) doing AccessTools.Field(...) repeatedly is an avoidable reflection overhead, so caching the FieldInfo once is a straightforward optimization (micro if you like to call it that way). no behavior change whatsoever, just less per tick cost

CSharpSourceCode/HarmonyPatches/PartyScreenPerformancePatches.cs
emrebahat: i will

CSharpSourceCode/HarmonyPatches/InventoryPatches.cs
Z3rca: Is this actually the case ? you are just accessing a static field that makes a lookup anyway. I would need to dive here deeper into things.

CSharpSourceCode/Extensions/ItemObjectExtensions.cs
Z3rca: well cache also means we are using more memory. we cant cache everything. we are now at somewhat 200+ traits and in future this is just more. not sure if this makes a difference on the long run, but we should be smart about those look ups. maybe in battles, we cache them, while outside they stay with accessors (since here it doesnt matter that much? that would increase slightly the load times, but wouldnt affect that much the memory usage (on the long run?)

CSharpSourceCode/HarmonyPatches/InventoryPatches.cs
emrebahat: yeah, AccessTools.Field(...) is executed once during the types static initialization. OnFrameTick doesnt call AccessTools, it only uses the cached FieldInfo.GetValue(), so repeated reflection name lookups per tick are avoided

Z3rca (approved):

PR #7, opened 2026-01-18
TORdiplomacy

Z3rca:
Allow for forming alliances and triggering diplomatic decisions for kingdoms (player and AI)
Key differences:

alliances work a little bit TWWH esque: when ally gets attacked , you have to solve a  "Honor the Allaince" request. which either let you join the ally in his war, or break the alliance with them. This should prevent a too passive gameplay.

grafik

Religion: changing from "allignment" to "Pantheon" for a more fine granular diplomacy. e.g Elves and dwarfs can now have "distrust" which makes alliances with them less likely. this was previously not possible. 

Nice side effect : Things i get constantly asked to "limit" or disable, are way easier to handle. There is in lore no occurence of humans praying to dwarf gods, or elfs to human gods. this makes these considerations easier. 

since Taleworlds did not add a logic to form alliances for AIs, i implemented a rudimentary basic behavior that does the trick.

Alliances , Trade Agreements and Honor Alliance considerations: follow a way more complex consideration. These are lore friendly and can account for multiple things, like personal relations, or even former grudges or "historic" liking or disliking.

Bandit population has been reduced significantly to keep up with the amount of wars that are not arising

Wars are back on the menu.

Everything had been tested on atleast 2 devices, working with same performance. as before. I am not sure if we slightly could increase the bandit population again (but not former level, that was too much).

Silence: 

With the LogEntryNotificationPatch, events are filtered that are not relevant to the player . this can be kingdoms , their distance, far distant heros etc. this makes the text in the "chat" a lot more readable and a lot more enjoyable.


CSharpSourceCode/CampaignMechanics/TORFactionDiscontinuationCampaignBehavior.cs
Z3rca, line 1: Request by Jojabo, to make Mousillon and Sylvania as well as Athel Loren and Eonir survive and join each other if they get defeated.

PR #9, opened 2026-02-26
Proper design changes

alemreM:
Faith Revival perk has been nerfed from 50% chance to 30%
Disabled vanilla campaign map events
Implemented a Sell & Donate all mechanic that stops when you hit the max gold/xp you can get
A WIP Custom Crush Through Block mechanic scaled with energy and skill difference, logging every calculation it makes
Fixed a bug(?) that caused lords to surrender when they joined a caravan/player party to help them against the player
Limited the amount of soldiers an army can leave a garrison to it's cap, avoiding constant deserts
Fixed garrison morale to 40 after it's capturement to avoid new owner's troops deserting the garrison
Fixed (?) or nerfed gear with rarity modifiers after a battle, now it's harder to find Epic ones and easier to find Uncommons
Reduced Warpstone drop rates by 3 times
Implemented an easily tweakable Retreat system to replace Vanilla's terrible pursuit and routing logic. 
Implemented a Lance-on-Foot logic for dismounted cavalry, this will hopefully prevent them from stabbing with lances when they are right in front of the enemy
Implemented village heart bonus multipliers for certain cultures, this will result in more prosperous towns, more food stocks, better protected villagers and richer merchants


CSharpSourceCode/BattleMechanics/TORBattleAgentLogic.cs
```diff
@@ -1,13 +1,88 @@
...
+                    out bool hasIncomingMountedCharge,
+                    out bool isIncomingMountedChargeTooCloseToSwapToLance);
+
+                bool allowLanceOnFoot = !hasCloseFootEnemy && hasIncomingMountedCharge;
+
+                if (allowLanceOnFoot)
+                {
+                    if (!isIncomingMountedChargeTooCloseToSwapToLance)
+                    {
+                        TryWieldSlotIfNotAlready(agent, lanceSlot.Value);
```
Z3rca, line 65: the terminology of the boolean is a bit a misleading : ALLOWING is not the same as enforcing. since this is what you do here.

CSharpSourceCode/BattleMechanics/TORBattleAgentLogic.cs
```diff
@@ -1,13 +1,88 @@
...
+
+            _elapsedSinceLastTick += dt;
+            if (_elapsedSinceLastTick < TICK_INTERVAL_SECONDS)
+            {
+                return;
+            }
+
+            _elapsedSinceLastTick = 0f;
+
+            foreach (var agent in Mission.Agents)
```
Z3rca, line 41: this sounds like an expensive agent tick. are we sure this is the best method to put this?

CSharpSourceCode/HarmonyPatches/CulturalFeatPatches.cs
```diff
@@ -0,0 +1,118 @@
...
+using TaleWorlds.CampaignSystem.Settlements;
+using TaleWorlds.CampaignSystem.Settlements.Buildings;
+using TaleWorlds.Core;
+using TaleWorlds.Localization;
+using TOR_Core.Utilities;
+
+namespace TOR_Core.HarmonyPatches
+{
+    // culture based hearth scaling WIP
+    [HarmonyPatch(typeof(DefaultSettlementProsperityModel), "CalculateHearthChangeInternal")]
```
Z3rca: i am not a fan of patching models. cant you just derive a TORSettlementProsperityModel from Default and then have the same?  I am not too close in the logic, i know sometimes those things are not easy to do.

we should avoid patches were we could, yet i know there are situations where it is cleaner to just patch it. just not sure if it is possible here.

Z3rca: > public override ExplainedNumber CalculateHearthChange(Village village, bool includeDescriptions = false)
>   {
>     ExplainedNumber result = new ExplainedNumber(includeDescriptions: includeDescriptions);
>     this.CalculateHearthChangeInternal(village, ref result, includeDescriptions);
>     return result;
>   }
> 

this seems for me like a good case of just creating an override, and call the base function. either before or after. no need for this patch. move this please in an approriate model

CSharpSourceCode/HarmonyPatches/EncounterPatches.cs
```diff
@@ -117,4 +118,37 @@ public static void GetPrisonerRosterPostfix(TroopRoster __result)
                 PendingLootedTroopManager.ApplyPrisonerModifications(__result);
         }
     }
+
+    [HarmonyPatch(typeof(DefaultEncounterModel), nameof(DefaultEncounterModel.GetSurrenderChance))] // villager parties and caravans will no longer surrender if there's a lord helping them
```
Z3rca: same as above, why do you need a patch here?

CSharpSourceCode/HarmonyPatches/FastTradeInventoryPatch.cs
```diff
@@ -0,0 +1,420 @@
...
+using TaleWorlds.CampaignSystem.Inventory;
+using TaleWorlds.CampaignSystem.Party;
+using TaleWorlds.CampaignSystem.Roster;
+using TaleWorlds.CampaignSystem.ViewModelCollection.Inventory;
+using TaleWorlds.Core;
+using TaleWorlds.Library;
+
+namespace TOR_Core.HarmonyPatches
+{
+    // Trading and Donating all stops when gold/xp thresold is reached
```
Z3rca: i am not sure what this means in context of this class name. patches need now a lot more description: Why did you make this patch, why was it maybe not possible to avoid it. so we can in future get rid of the patch or understand why its redundant etc.

CSharpSourceCode/HarmonyPatches/TradeSkillPatch.cs
```diff
@@ -0,0 +1,121 @@
...
+        private const float EQUIPMENT_TRADE_SKILL_BONUS_PER_LEVEL = 0.0073f; // same prices as vanilla at 300 trade
+        private const string TRADE_PENALTY_REDUCTION_DESCRIPTION =
+    "{=str_tor_trade_skill_description}Trade penalty Reduction +0.2% per level for trade goods and animals.\n" +
+    "Trade penalty Reduction +0.7% per level for equipment.";
+
+        [HarmonyPatch(typeof(DefaultTradeItemPriceFactorModel), nameof(DefaultTradeItemPriceFactorModel.GetTradePenalty))]
+        private static class EquipmentTradePenaltyUndoPatch
+        {
+            [HarmonyPostfix]
+            private static void Postfix(
```
Z3rca: again, this is a very simple Method override. why is this a harmony patch.

CSharpSourceCode/HarmonyPatches/TradeSkillPatch.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            CharacterObject partyLeader = SkillHelper.GetEffectivePartyLeaderForSkill(clientParty.Party);
+            return partyLeader?.GetSkillValue(DefaultSkills.Trade) ?? 0;
+        }
+
+        [HarmonyPatch(typeof(SkillHelper), nameof(SkillHelper.GetEffectDescriptionForSkillLevel))]
+        private static class TradePenaltyReductionDescriptionPatch
+        {
+            [HarmonyPostfix]
+            private static void Postfix(SkillEffect effect, int level, ref TextObject __result)
```
Z3rca: all methods here are postfix, this cries for a simple override of the method.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -979,35 +980,198 @@ public override float CalculateRemainingMomentum(float originalMomentum, in Blow
             return newMomentum;
         }
 
+        private const int EXTRA_CTB_SKILL_DEAD_ZONE = 30; //skill diff over 30 doesn't do ctb
```
Z3rca: why are the constants not closer to the class definition. Also , i get why Crush Through Block is not easy to write, and for a variable i think its fine , then atleast write a heading over a block of these constants : 

 //Crush Through Bock (CTB)
... constants...

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -979,35 +980,198 @@ public override float CalculateRemainingMomentum(float originalMomentum, in Blow
...
+        private const float EXTRA_CTB_NON_OVERHEAD_CHANCE_MULTIPLIER = 0.5f; // non overhead swings are twice as less likely
+        private const float EXTRA_CTB_SKILL_GROWTH_RATE = 0.008097f; // 30 energy, 50 diff 50% ctb, 200 diff 50% ctb
+
+        // how quickly extra ctb reaches full effect after passing the base energy threshold
+        // reduce this number if you want momentum to matter less on ctb
+        private const float EXTRA_CTB_ENERGY_MARGIN_FOR_FULL_EFFECT = 0.27f;
+
+        private static readonly float ExtraCtbSkillGrowthRate = EXTRA_CTB_SKILL_GROWTH_RATE;
+
+        private static float CalculateExtraCtbSkillGrowthRate()
```
Z3rca, line 1008: no fan of having method names  not containing the full name though (see above) this is especially confusing without a method description. so either add a method description, or use a method name without abbreviation.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -979,35 +980,198 @@ public override float CalculateRemainingMomentum(float originalMomentum, in Blow
...
+            }
+
+            float normalized = numerator / denominator;
+            float chance = EXTRA_CTB_TARGET_OVERHEAD_CHANCE_AT_TARGET_DIFF * normalized;
+
+            return MBMath.ClampFloat(chance, 0f, 1f);
+        }
+
+
+        private static float CalculateExtraCtbEnergyFactor(float totalAttackEnergy, float threshold)
```
Z3rca, line 1051: Overall CrushThroughBlock needs more description i feel.

Z3rca (changes requested):
work on the comments, the logic seems fine and i have no issues with the desired changes. the only thing i would want to check is the way you enforce the spear wielding. I do have problems though with how these changes are implemented. Patching should always be the last resort, and needs a lot description, so we are not falling in the traps of the past of having no idea what patches do, or if they are still needed (after mayor updates). So avoid this as much as possible. CTB i already told, needs a bit more description otherwise I lose my mind reading some abreviation in the future, that is not trivial to understand.

Z3rca (approved):

PR #12, opened 2026-03-08
Naber various and minor fixes

alemreM:
WIP stutter on order optimization
Morrslieb ink fix for fast mode

allied greenskin troops affect shaman ability
Catalyst perk and rewield weapons fix
and greenskin mercenary but that's not really important


CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
```diff
@@ -477,6 +502,12 @@ private void HandleAnimations()
 
         private void UpdateWieldedItems()
         {
+            if (Agent.Main == null || !Agent.Main.IsActive())
+            {
+                ClearMainAgentAbilityStateAfterRemoval();
+                return;
+            }
```
Z3rca, line 509: did you test this also for Off hand staffs?

CSharpSourceCode/Models/TORAgentStatCalculateModel.cs
```diff
@@ -419,178 +419,196 @@ private void UpdateAgentDrivenProperties(Agent agent, AgentDrivenProperties agen
...
                     }
                 }
+
                 if (!agent.WieldedWeapon.IsEmpty)
                 {
                     if (agent.WieldedWeapon.Item.IsMeleeWeapon())
                     {
-                        var traits = agent.WieldedWeapon.Item.GetTraits().WhereQ(x => x.StatsTuple?.StatType == ItemTraitStatType.SwingSpeed).ToListQ();
-                        foreach (var trait in traits)
+                        foreach (var trait in agent.WieldedWeapon.Item.GetTraits())
```
Z3rca, line 435: those WhereQ and ListQ are queries that improve access time (aparently, never tested the actual performance gain. These operations here are very frequently used, so I thought performance is key here. what is the gain not doing it ?

Z3rca: hm you might be right here. however, just as a test: try to make GetTraits to a MBList, and see if WhereQ does actually improve performance. those "Queries" wont work on regular list items well and will create actually overhead. however TW introduced them to improve performance. So as a test maybe try to make the List<T>, as MBList<T> items

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
emrebahat: they should be exempt from it since they are not sheathed in the first place

CSharpSourceCode/Models/TORAgentStatCalculateModel.cs
emrebahat: GetTraits() already returns a fresh built list so WhereQ/ToListQ feels like extra work on top of an already materialized list

PR #15, opened 2026-05-11
Feature/runesmith shop refactor

SlyDevil:
Pull request on behalf of Randy for runesmith shop changes.

They can comment further.


RandyChihuahua:
Notes taken from Discord:

I've added shields
Added Armour based on a list of items (as tier is not doable) to lower tier runesmith guild unlocks.

- Tier 1 unlocks dwarf warrior, militia and miner gear
- Tier 2 unlocks longbeard armour + shield
- Tier 3 remains the same.
- Klad Brakak is no longer sold, as @Pierogolepca mentioned that it should not be.
- String lists abstracted to their own class, (this could PROBABLY be abstracted to XML - but I would rather this be merged if accepted, then we can work on the TODO's)
- I've increased the quantity of buying for all items, it doesn't matter as you can go in and out of the menu for more anyways.
- Refactored the code to be slightly more efficient.

SlyDevil:
I have no particular concerns with the code aspects.
There's a potential issue I see, but it's in the original code before your changes so 🤷 , not immediately relevant.

With, f.e., Runesmith 2 attribute check, the prior filter was for item tiers `< Tier 4`, ie. T3 items and below, and was changed to `<= Tier 5`; were there specific items you were looking to have included when increasing the viable item tier by 2?

RandyChihuahua:
Can always fix the potential issue? 

For the tiers I wanted to avoid having a set list of shields to include in the list, so I went with the .IsShield() method and allowed tier 5 as the longbeard shield is tier 5.

3 potential options here:

1. Reduce the longbeard shield tier manually, wasn't sure where to do this and what other impact it could have
2. Remove the .IsShield() method and add it manually to the tier 2 unlock
3. Keep it as is. I chose this option because weapons aren't really a problem to obtain or unlock, given the fact you can smith them so easily yourself, it's not unlocking ranged or unique weapons.

Feedback is also best left on the lines themselves ;)
image

SlyDevil:
The runesmith guild is already the most powerful guild for player power increases due to enchantment access as well as irondrake gear. We're not looking to reinforce that by adding direct faction gear access to their shop.

For now we'll leave access to that equipment through town production via workshops.

Having more reliable cultural equipment access for the player is generally of interest, but that would probably be in a more specific system that could be applied to all of the cultures rather than each one needing a specific implementation.

PR #21, opened 2026-06-09
Refactor for OnSettlementEntered events for custom settlements

SlyDevil:


CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
```diff
@@ -543,99 +544,26 @@ private void OnNewGameStart(CampaignGameStarter starter)
...
-                    else
-                    {
-                        _lastGhostRecruitmentTime.Add(party.LeaderHero.StringId, (int)CampaignTime.Now.ToDays);
-                    }
-                }
-            }
-        }
+        if (party == null || leaderHero == null || party == MobileParty.MainParty) return;
 
         LeaveSettlementAction.ApplyForParty(party);
```
SlyDevil, line 578: What triggers first, the component's OnPartyEntered, or the generic OnSettlementEntered event?

Is this safe, or could the LeaveSettlementAction be performed before all of the mechanics having occurred which depend on the party's CurrentSettlement being non-null?

SlyDevil:
What sorts of tests have you done for verifying that no functionality is broken for the various settlement components?

SlyDevil:
I see no evident logic errors which need fixing.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
RandyChihuahua: The OnPartyEntered is activated first, this has all been tested and works fine.

On a side-note, i tried as much as I could to remove the need for a OnSettlementEntered event registry, unfortunately it is still mandatory, but I have moved as much of the logic as I can into the OnPartyEntered.

RandyChihuahua:
> What sorts of tests have you done for verifying that no functionality is broken for the various settlement components?

Live testing; since component and unit tests do not exist. 

Dwarf lords will enter the grimnir settlement, gain slayers and leave as required.

Vampire counts will enter the settlement, gain ghosts and leave as required.

CSharpSourceCode/CampaignMechanics/TORCustomSettlement/TORCustomSettlementCampaignBehavior.cs
SlyDevil: Hmm, it seems odd to me that having the party ai recalculate it's next decision doesn't allow it to make a new decision that leads to it triggering its own LeaveSettlementAction on a future tick.

A party can enter a town, wait a couple hours, then it chooses a new target and exits the town of its own volition. I would assume that the same behaviour could be triggered for these parties without forcing an exit.

SlyDevil:
I like the shift to keeping the behaviour in a single location as the OnPartyEntered event didn't pan out from your testing.
The namespace organization is better with each component being independent.
Code is more readable with better escaping and less indentation.

SlyDevil:
Recompiling of the dll will be handled later when I deal with other commits.

PR #24, opened 2026-07-31
Bugfix - Clans were not joining kingdoms due to a rounding integer bug:

RandyChihuahua:
The chance for someone to join is 10% divided by the number of heroes * 2. 

See line 104, it has an integer type/ Unfortunately the moment you go over 10 (10 / 11 for example), you'll have less than 1, which rounds to 0. I have changed this to a float.

I've done some refactoring as well. Line wrapping and such.


Z3rca:
had at first problems understanding the problem since we used var, but yes that is indeed a bug thanks for fixing.

For future : maybe add in code quotation marks (github issues has something like this the exact code change. the shifting around of code was a bit confusing to me.

Z3rca (approved):

RandyChihuahua:
Sly has requested testing on the matter. Will do that before merging.

RandyChihuahua:
image
Moved the Melkior settlement to my own, commented out the 30% part, and got this result. It's now properly evaluating as a float value.

PR #26, opened 2026-07-31
Visual studios -> Remove and sort usings.

RandyChihuahua:
You can do this in future by going to visual studios' code cleanup function.

No AI jazz, it's builtin and should be done at least once.


Z3rca:
looks like good idea, thanks for looking into that. i always saw those millions of stale imports, that had absolutely no use. is properly a good idea to add that

Z3rca (approved):

PR #27, opened 2026-08-02
replace NAudio with engine audio

emrebahat:


bin/Win64_Shipping_wEditor/TOR_Core.dll
Z3rca, line 1: ist this right, dont we need a editor dll ?

Z3rca (approved):
wait for a go of hunharibo, i am fine with this from what i understood. if you can test, please make sure you discharge audio sources from memory, which is the biggest culprit for sounds

PR #28, opened 2026-08-29
Maintenance/code smells

RandyChihuahua:
Various code smells, please note that I wasn't able to fix them all directly, as they require some context around either the battle engine or otherwise.


CSharpSourceCode/BattleMechanics/Artillery/FieldTrebuchet.cs
```diff
@@ -119,7 +119,18 @@ protected override void OnInit()
 			list = base.GameEntity.CollectScriptComponentsWithTagIncludingChildrenRecursive<SynchedMissionObject>("rotate_entity");
 			this.RotationObject = list[0];
 			base.OnInit();
+			// PROPOSED (CS1717, self-assignment -- possible real bug, not just lint noise): this
```
SlyDevil: It's wholly redundant. The field that's being written to is already accessible because we derive from its class and it's `protected.`
Changing to a `const` declaration to hold the value in code, and setting the field to its value during OnInit.

CSharpSourceCode/BattleMechanics/Dismemberment/DismembermentMissionLogic.cs
```diff
@@ -31,7 +31,11 @@ public class DismembermentMissionLogic : MissionLogic
         //14*20 = 280 Game Entities are spawned. keep the pool size as reasonable small as possible.
         private GameEntity[][] _pooledDismemberedLimbs;
         private int _index;
+        // PROPOSED (CS0414, assigned once below, never read): a search of this file found no reader.
+        // RANDY - @Sly i'll let you make the call there.
+#pragma warning disable CS0414
         private bool _fullyInstantiated;
```
SlyDevil: I'd assuming testing variable left in. Removed.

CSharpSourceCode/BattleMechanics/TriggeredEffect/TriggeredEffect.cs
```diff
@@ -16,7 +16,13 @@ namespace TOR_Core.BattleMechanics.TriggeredEffect
     public class TriggeredEffect(TriggeredEffectTemplate template, bool isTemplateMutated = false) : IDisposable
     {
         private TriggeredEffectTemplate _template = template;
+        // PROPOSED (CS0414, assigned -1 in two places below, never read): PlaySound uses its own
+        // local `soundIndex` now, so this field looks like a leftover from before sound playback
+        // moved to the queued AbilityManagerMissionLogic.QueueTriggeredEffectSound path.
+        // RANDY - @Sly i'm not familiar enough with the engine to know if this impacts anything.
```
SlyDevil: Native sound manager handles its own ids without intervention.
It sets ids to -1 to allow it to identify which are waiting for disposal.

CSharpSourceCode/BattleMechanics/CinematicCameraMissionView.cs
```diff
@@ -14,7 +14,13 @@ public class CinematicCameraMissionView : MissionView
         private bool _isCameraActive = false;
         private float _cameraSpeed = 5f;
         private MissionMainAgentController _agentController;
+        // PROPOSED (CS0108, should hide with `new` or `override`): `MissionView` already declares
```
SlyDevil: MissionView uses `MissionScreen.SceneLayer.Input` which restricts inputs to whatever is accepted by the current layer, eg. pressing a f# key in battle will add the formation commands scene layer which filters how key presses are then accepted/interpreted.
Using the global input manager seeks the key pressed with no context filtering.

Changed to `new`.

If for some reason this class is inherited from in the future, we can review whether the access modifier should be readded.

CSharpSourceCode/CampaignMechanics/UniqueSpawns/UniqueSpawnPartyComponent.cs
```diff
@@ -31,12 +31,9 @@ public class UniqueSpawnPartyComponent : WarPartyComponent
         private string _uniqueSpawnId;
 
         [SaveableField(7)]
-        private bool _consumesFood; // old saves
-
-        [SaveableField(8)]
         private int _startingFoodPerType;
```
SlyDevil: Food for these parties should be wholly irrelevant, they're set to not consume any food in the Consumption model.
I think both of these fields were from an earlier version during initial development and are unnecessary.

@emrebahat Is there a reason this party component needs 100 of the like 12 food types?
If we're introducing a save incompatibility by removing one, may as well remove the other at the same time.

CSharpSourceCode/BattleMechanics/StatusEffect/StatusEffectComponent.cs
```diff
@@ -234,7 +234,15 @@ private void CalculateEffectAggregate()
             }
         }
 
+        // PROPOSED (CS0114, should add `new` or `override`): StatusEffectMissionLogic.cs calls this
+        // through the concrete `StatusEffectComponent` type (not polymorphically via
+        // `AgentComponent`), so `new` would match current behavior. Before adding `override` instead,
+        // worth checking whether the engine also auto-invokes AgentComponent.OnTick on every
```
SlyDevil: Added a clarifying remark for why we can't use the OnTick override that's available via AgentComponent and renamed the method to better differentiate between the seeming "Tick" calls.

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/CareerChoicesHelper.cs
```diff
@@ -33,10 +33,9 @@ public static bool HealthLostCheck(Agent agent, float healthLost)
             return agent.HealthLimit - agent.Health >= healthLost;
         }
 
-
         public static bool ContainsSpellType(AbilityComponent component, int spellCount, AbilityEffectType excludedEffectType)
         {
-            var wrongSpell = false;
+            //TODO: Why not use linq?
```
SlyDevil, line 38: I once stumbled upon a comment from TW, I think, which recommended using indexed for loops instead of foreach loops f.e. because the former were more performant.

There wasn't much context on the circumstances in which they were more performant, but this may be an implementation of a similar thought process.

Idk if there's any performance benefit or cost to LINQ, but it would probably be slightly easier to read intention.

On the otherhand, I'm thoroughly confused as to why `GetAbility(i)` performs a modulo on the passed index as having circular indexing seems useless when the iterations are all limiting themselves to `.Count`.
There's derivative uses that don't check count, but they're pulling the useable AbilityTemplates from the hero's selected ability list in which case the index must necessarily exist within the bounds of the list.

Rabbit hole for another day when the ability templates get a rewrite.

SlyDevil:
I've changed everything relevant and added clarifications where useful.

I'll let you handle the merge conflict. Afaict, it's to override the next_update EnchantmentHelper contents with the contents of this branch, but it's too late for the brain to be confident in that.

SlyDevil:
Branch is ready for merge.

Leaving some time for naber to respond about the party food, but from what I see it can be removed completely along with usages of it.
If the parties need food items, that default can be put in a constant or the argument passed elsewhere. I don't see a reason to need the integer saved to file when it's only used on party creation.

CSharpSourceCode/CampaignMechanics/UniqueSpawns/UniqueSpawnPartyComponent.cs
emrebahat: no not anymore. from an early prototype.

PR #29, opened 2026-08-30
Refactor/enchantment helpers

RandyChihuahua:
Mostly splitting into 2 files, and refactoring one completely. I tested and it works.

Also added a condition when the player has neither enough gold or special currency.


CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        var lore = LoreObject.GetAll().FirstOrDefault(x => x.StringId == restriction);
+        return lore != null
+            ? "This enchantment is bound to the Lore of " + lore.Name + ". "
+            : "This enchantment requires " + restriction + ". ";
+    }
+
+    private static int CalculateCustomResourceCost(int skillValue)
+    {
+        var factor = Hero.MainHero.GetCultureSpecificCustomResource().GetCustomResourceGeneralizedFactor();
```
SlyDevil, line 171: I'm unsure to what degree that factor has been reflected upon, balanced around, or tested. This will need specific testing to see how it plays out in practice.

docs/vertical-slicing-proposal.md
```diff
@@ -0,0 +1,251 @@
...
+| `AbilitySystem/Scripts/` | **SPLIT** | Base `AbilityScript` machinery → FW. Per-Career `CareerAbilityScript` subclasses → `Modules/Careers/Abilities/`. |
+| `BattleMechanics/StatusEffect/`, `TriggeredEffect/`, `DamageSystem/` | FW | The shared effect-resolution pipeline every spell/item/prayer fires through. |
+| `BattleMechanics/AI/` (all of it: `CastingAI/`, `TeamAI/`, `ArtilleryAI/`, `CivilianMissionAI/`, `CommonAIFunctions/`) | FW | Battle-simulation infrastructure, not a toggleable feature — every battle uses it regardless of which content modules are involved. |
+| `BattleMechanics/` root (`TORBattleAgentLogic`, `AddAgentComponentsMissionLogic`, `CustomCrosshairMissionBehavior`, `CinematicCameraMissionView`) | FW | Generic mission plumbing. |
+| `BattleMechanics/` root — `CareerPerkMissionBehavior` | MOD → `Careers/` | Career-specific despite living at `BattleMechanics/` root today. |
+| `BattleMechanics/` root — `TORMonsterSiegeLogic`, `SiegeEarlyVictoryMissionLogic` | MOD → `TORCustomSettlement/` | Monster-siege support exists for Troll Cave content. |
```
SlyDevil: Siege support is for things like troll and treemen being able to bash down gates and not try to climb stairs as they'll get stuck due to size.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
RandyChihuahua: Ideally this will all behave the same way as before, we just need some manual testing that we don't get any weird behaviors. This is where testing expertise comes in. When I loaded up a game as a runesmith though, it enumerated all the menus and created a Runesmith character fine.

docs/vertical-slicing-proposal.md
RandyChihuahua: Okay, i've updated that line manually. Good catch.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -63,36 +64,22 @@ private static List<PurchasableBlueprint> GetPurchasableBlueprints(List<string>
...
-        var enabled = eligibleHeroes.Any(hero => hero.GetSkillValue(skill) >= skillValue);
+        var enabled = blueprint.EligibleHeroes.Any(hero => hero.GetSkillValue(blueprint.RequiredSkill) >= blueprint.RequiredSkillValue);
 
         var hintText = new TextObject("{TRAIT_EFFECT}\n\n{REQUIREMENT_TEXT}\n\n{COMPLETE_COST}");
-        hintText.SetTextVariable("REQUIREMENT_TEXT", enabled ? "" : BuildRequirementText(eligibleHeroes, skill, skillValue, blueprint.Restriction));
+        hintText.SetTextVariable("REQUIREMENT_TEXT", enabled ? "" : BuildRequirementText(blueprint.EligibleHeroes, blueprint.RequiredSkill, blueprint.RequiredSkillValue, blueprint.Restriction));
 
-        var crCost = CalculateCustomResourceCost(skillValue);
-        var goldCost = item.Value;
+        var crCost = CalculateCustomResourceCost(blueprint.RequiredSkillValue);
```
SlyDevil, line 79: Hmm?
Why is a custom resource cost needing a skill value argument? No skills have passive effects related to enchanting in general, nor specifically cost of purchasing new enchantments from a trainer.
I would expect the cost to be invariable wrt a skill's level.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -106,9 +93,9 @@ private static InquiryElement CreateInquiryElement(PurchasableBlueprint blueprin
         hintText.SetTextVariable("COMPLETE_COST", "{GOLD_VALUE}{GOLD_ICON} , {CR_VALUE}{CUSTOMRESOURCE}");
         GameTexts.SetVariable("CR_VALUE", crCost);
         GameTexts.SetVariable("CUSTOMRESOURCE", Hero.MainHero.GetCultureSpecificCustomResource().GetCustomResourceIconAsText());
```
SlyDevil, line 93: Fyi, if the size of the icon looks off in the description, there's a method for fetching the icon that takes an argument for scaling allowing adjustment for a specific text context.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
SlyDevil: So the skill value on a blueprint is being used as the base cost when learning it.
Ppf, works.
If that's not already documented, can you add that to RequiredSkillValue so it's more evident that the value is doing double duty beyond what its name implies.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -21,10 +21,26 @@ namespace TOR_Core.CampaignMechanics.Crafting;
 public static class EnchantmentShopHelper
 {
     public static void OpenEnchantmentRecipeShop(List<string> prefixList, string culture, bool blessings = false)
+    {
+        var purchasableBlueprints = GetPurchasableBlueprints(prefixList);
+        var selectableItems = BuildInquiryElements(purchasableBlueprints);
+
+        var shopVariation = GetShopVariation(culture, blessings);
+        var title = GameTexts.FindText("tor_enchantmentshop_title", shopVariation).ToString();
```
SlyDevil: Can use TORTextHelper which extends FindText to also include default strings and such.
Its goal is to allow game designers to adjust strings without needing to touch any code.

When developing, we only need to choose the string id, but the string contents can be updated later by someone else. Particularly useful for letting more lore knowledgeable people write descriptions, dialogue, etc...

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -37,166 +53,167 @@ public static void OpenEnchantmentRecipeShop(List<string> prefixList, string cul
...
-                    requirementPrefix = "This enchantment is bound to the Lore of " + lore.Name + ". ";
-                }
-                else
-                {
-                    requirementPrefix = "This enchantment requires " + restriction + ". ";
-                }
-            }
+        if (enabled)
+        {
+            hintText = new TextObject(trait.ItemTraitDescription + "\n {GOLD_VALUE}{GOLD_ICON} , {CR_VALUE}{CUSTOMRESOURCE},\n {VALIDTYPE_RESTRICTION}");
```
SlyDevil: Use `{newline}` when making a line break. The game replaces it with `\n` at runtime, but when a string is pulled from an xml (ie. localized text replacement), the `\n` has issues.

Is your intention to have a line break, then offset the text on that new line by one space : `\n(space){TEXT_VARIABLE}`? ie. all of the additional lines after the trait description are indented by 1 space.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
```diff
@@ -37,166 +53,167 @@ public static void OpenEnchantmentRecipeShop(List<string> prefixList, string cul
...
+        }
 
-            if (underlyingTrait != null)
-            {
-                string typeRestriction = GameTexts.FindText("tor_enchantmentshop_restriction", underlyingTrait.ValidItemType.ToString()).ToString();
-                GameTexts.SetVariable("VALIDTYPE_RESTRICTION", typeRestriction);
-            }
+        var lore = LoreObject.GetAll().FirstOrDefault(x => x.StringId == restriction);
+        return lore != null
+            ? "This enchantment is bound to the Lore of " + lore.Name + ". "
```
SlyDevil: This is an example for where to use the TORTextHelper.
You can define a generic description that contains all of the text variables used in code, but the exact contents can be left to game design. As long as you take the TextObject and set its variables, whatever is written after in the xml will function correctly without further input.

Strings necessarily need ids and to pass through TextObjects for them to be available to localization. See [the wiki page](https://github.com/TheOldRealms/TOR_Core/wiki/String-and-localization-ID-nomenclature) for additional details. (The intrapage hyperlinks are broken currently, they're supposed to jump to section headers.)

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
SlyDevil: I see what you mean.
This factor was already used in the same calculation; I was thinking about an old form of it.

RandyChihuahua: Going to close this comment for now.

CSharpSourceCode/INDEX.md
```diff
@@ -0,0 +1,134 @@
...
+
+- **`CampaignBehaviorBase` per mechanic** — almost every gameplay system in
+  `CampaignMechanics/` is one behavior class registered in `SubModule.InitializeGameStarter`.
+  To find how a mechanic starts, grep its behavior class name in `SubModule.cs`.
+- **`GameModel` overrides** — `Models/` replaces vanilla formulas one at a time
+  (`TORXyzModel : DefaultXyzModel`), registered in `SubModule.OnGameStart`. To change a
+  formula, find the matching model here before writing a Harmony patch.
+- **Harmony patches** (`HarmonyPatches/`) are the fallback for anything vanilla doesn't
+  expose a model/behavior/virtual method for. Most patch at `OnSubModuleLoad`; a few need
+  `[HarmonyPatchCategory("LatePatches")]` to run after `Game.Current`'s text manager exists.
```
SlyDevil: This is an incomplete reason.

Harmony patching a class causes the constructor to run. If you cause it to build too early, it initializes static members with default values which are then never updated and you get knock-on effects. [example](https://docs.bannerlordmodding.lt/guides/human_bullet/)

Patches that try to fetch things from managers or other fields also need to wait for them to be initialized, but these at least are generally direct crashes that are easy to identify.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
RandyChihuahua: Done.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
RandyChihuahua: Done.

CSharpSourceCode/INDEX.md
RandyChihuahua: Thanks for the headsup. I've removed the index.MD's from the file for now.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
RandyChihuahua: I added a comment:
//Note that RequiredSkillValue is not only gating purchasability, but it's also setting the cost for custom resources.

CSharpSourceCode/CampaignMechanics/Crafting/EnchantmentShopHelper.cs
RandyChihuahua: Will address that in MR on centralizing purchase of enchantments.

PR #30, opened 2026-09-06
Raiding clan removal

RandyChihuahua:


ModuleData/tor_clans.xml
```diff
@@ -1312,13 +1312,20 @@
...
+             label_color="FF8B7C73"
+             color="FF8B7C73"
+             color2="FF8B7C73"
+             alternative_color="FF8B7C73"
+             alternative_color2="FF8B7C73"
+             text="{=str_tor_clan_chaos_clan_1_desc}Chaos is the term used by mortals to describe the malign spiritual force of supernatural disorder that afflicts the Known World. It is a potent force composed of pure magic that can unleash physical mutation and spiritual corruption upon sentient beings. In perspective, Chaos is akin to the ocean, a constant, roiling force of emotion-fuelled change, the very opposite of the rigid physical laws that govern the natural world. Like the ocean, its power is one that waxes and wanes over time, but this tide constantly batters and seeks to invade the mortal plane of existence.">
+    </Faction>
     <Faction id="beastmen_clan_1"
              name="{=str_tor_clan_beastmen_clan_1_name}Beastmen"
-             initial_home_settlement="Settlement.herdstone_01"
```
RandyChihuahua, line 1317: Can this "beastmen_clan_1" not be utilized with the default bandit parties above?

SlyDevil: Possibly, in that we create RaidingPartyComponents and handle their ai separately from BanditPartyComponents so the "steppe_bandits" beastmen clan being marked as "is_bandit" wouldn't directly be problematic.
But there's hardcoding related to the native bandit clan ids (steppe_bandits, forest_bandits, etc...) which is why those are kept around and their ids reused for some of our bandit clans.

You'll also notice that the culture the bandit clan points at is a bandit culture with its own attributes that differ from playable and kingdom cultures. With beastmen being introduced as an entirely playable race, a beastman culture will be created and "beastmen_clan_1" will likely point at it instead of the "chaos_culture".

RandyChihuahua: Yeah, we'll leave it, it works.

SlyDevil:
This branch is being left to the side for the moment.

The IsBanditClan check is too prevalent throughout native code, requiring extensive patching and workarounds to have this function.

A request has been made to Taleworlds to create space for a new supercategory of heroless clans of which bandit clans are a subcategory. We could then have behaviors which apply to specific clan types and the native code would be more robust with regards to modded clan types being added that deviate from native ones.

See https://discord.com/channels/411286129317249035/677511186295685150/1547297308595454102 for the request and explanation provided for its benefits and the current limitations.

PR #31, opened 2026-09-07
Ironbreaker career perk/ability fixes

Dannyboi802:
- Correct troop targeting, ammunition bonuses, and upgrade discounts.
- Fix shield-block charging, Shieldwall targeting, and explosion/friendly-fire defenses.
- Increase Impenetrable duration and separate Gromril/Rune perk activation.
- Update descriptions.


CSharpSourceCode/BattleMechanics/DamageSystem/TORDamageHelper.cs
```diff
@@ -14,6 +16,84 @@ namespace TOR_Core.BattleMechanics.DamageSystem
...
     public static class TORDamageHelper
     {
+        /// <summary>
+        /// Halves the remaining friendly-fire damage after every other modifier.
+        /// Native hits and already-calculated scripted hits call this at their respective final damage boundaries.
+        /// </summary>
+        public static float ApplyIronbreakerFriendlyFireReduction(Agent attacker, Agent victim, float damage)
+        {
+            if (Campaign.Current == null ||
+                Hero.MainHero?.HasCareerChoice("GromrilArmorPassive4") != true ||
```
RandyChihuahua, line 26: Could pattern matching in conjunction with a NOT apply to any of these? might simplify it a bit.

!(attacker.Team is Team.Valid) for example.. food for thought. Close if wrong.

I like the guard clauses though.

CSharpSourceCode/CharacterDevelopment/CareerAbilityChargeSupplier.cs
```diff
@@ -507,11 +507,12 @@ public static float KnightOldWorldChargeSupplier(Agent affectingAgent, Agent aff
         public static float IronbreakerChargeSupplier(Agent affectingAgent, Agent affectedAgent, ChargeType chargeType, int chargeValue,
             AttackTypeMask mask = AttackTypeMask.Melee, CareerHelper.ChargeCollisionFlag collisionFlag = CareerHelper.ChargeCollisionFlag.None)
         {
-            // Damage taken: check if the affected agent (one being hit) is the main hero - 5x charge from blocking
-            //Sly : this doesn't account for blocked hits in any way?
+            // Damage taken grants 5x charge; shield blocks grant a fixed 5 even when no health is lost.
             if (chargeType == ChargeType.DamageTaken && affectedAgent == Agent.Main && affectedAgent.GetHero() == Hero.MainHero)
             {
-                return chargeValue * 5;
+                return (collisionFlag & CareerHelper.ChargeCollisionFlag.HitShield) != 0
```
SlyDevil, line 513: Should this be shield only, or the enum is expanded to include weapon blocks and any sort of blocking counts?

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -1186,6 +1186,12 @@ public float CalculateWardSaveFactor(Agent attacker, Agent victim, float[] resis
             {
                 result.LimitMin(0.11f);
```
SlyDevil, line 1187: Limit the min to 0.1f so an ironbreaker can actually benefit from the 90% ward save listed instead of 89%.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -1186,6 +1186,12 @@ public float CalculateWardSaveFactor(Agent attacker, Agent victim, float[] resis
...
             }
+            else if (victim.HasAttribute("Impenetrable"))
```
SlyDevil, line 1189: Add attributes to TORConstants.CharacterAttibutes. Their spelling only needs to exist in 1 location to limit errors.

CSharpSourceCode/Models/TORAgentStatCalculateModel.cs
```diff
@@ -155,36 +164,24 @@ public override void InitializeMissionEquipment(Agent agent)
...
+                                        var ironbeardCount = MobileParty.MainParty.MemberRoster.GetTroopRoster()
+                                            .Where(elem => elem.Character.StringId == "tor_dw_ironbeard")
+                                            .Sum(elem => elem.Number);
+                                        if (ironbeardCount > 0)
                                         {
-                                            if (elem.Character.StringId == "tor_dw_ironbeard")
-                                            {
-                                                ammoCount.AddFactor(0.1f);
-                                            }
+                                            ammoCount.AddFactor(0.01f * ironbeardCount);
```
SlyDevil, line 176: Store the 1% value in the PassiveEffect on the career choice, then fetch the choice and use `choice.GetPassiveValue()`.

CSharpSourceCode/CharacterDevelopment/CareerAbilityChargeSupplier.cs
Dannyboi802: it should be the shield only. ill test it

SlyDevil: The question is about design intent, not functionality.
The restricted mechanical space is self-imposed. We can change the underlying flags to permit detecting other forms of defending oneself beyond blocking with a shield.

If shield-only blocks are the intention, the career ability description should probably specify that.

CSharpSourceCode/Models/TORAgentStatCalculateModel.cs
```diff
@@ -155,36 +164,24 @@ public override void InitializeMissionEquipment(Agent agent)
...
-
-                                        if (missionWeapon.Item.IsFlameThrowerItem())
-                                        {
-                                            if (Hero.MainHero.HasCareerChoice("IronDrakesPassive3"))
-                                            {
-                                                ammoCount.Add(12);
-                                            }
-                                        }
-
+                                        ammoCount.Add(12);
```
SlyDevil, line 184: Store specific values where possible on the choice so description and value are defined in the same place.

There's probably other lines this applies to, but not going to leave comments on them all.

CSharpSourceCode/Models/TORPartyTroopUpgradeModel.cs
```diff
@@ -17,8 +17,10 @@ public override ExplainedNumber GetGoldCostForUpgrade(PartyBase party, Character
             if (characterObject.IsUndead()) return new ExplainedNumber(0);
 
             var explainedNumber = base.GetGoldCostForUpgrade(party, characterObject, upgradeTarget);
+            var applyIronbreakerDiscountLast = party.LeaderHero?.HasCareerChoice("IronPricePassive3") == true &&
```
SlyDevil, line 18: Use an early out here for all non-MainParties as all subsequent evaluations don't matter to them.
```suggestion
            if (party.LeaderHero == null || party != PartyBase.MainParty || party.LeaderHero != Hero.MainHero) return explainedNumber;
```

CSharpSourceCode/Models/TORPartyTroopUpgradeModel.cs
SlyDevil, line 60: Change this to an Add instead of AddFactor.
.Add() will affect the base number that all factors are applied on and prevent this +300% increase from being additive with other factors. Consequently, all of the additions here for the IronPricePassive3 perk can be removed.
```suggestion
                        explainedNumber.Add(explainedNumber.BaseNumber * 3);
```

This won't be exactly identical to what's written below due to `base.GetGoldCostForUpgrade` factors that would have been folded into the BaseNumber and would therefore be multiplicative with the IronPricePassive3 reduction, but it will be relatively similar, simpler, and more consistent with how factors are applied in the game.

Few factors are multiplicative; making one so should have a conscious reason for why it deviates and description formulations considered in how we communicate to the player that a value is being treated differently from others.

SlyDevil, line 28: This .MainParty check is superfluous if the guard clause is added higher up and can be removed.
`party.LeaderHero...` covers both the check for MainParty and has the incidental effect of checking if the player is a prisoner and would be able to attempt an upgrade in that state.

CSharpSourceCode/Extensions/CharacterObjectExtensions.cs
```diff
@@ -422,10 +422,11 @@ public static Tuple<CustomResource, int> GetCustomResourceRequiredForUpgrade(thi
                     var explainedNumber = new ExplainedNumber(cost);
                     CareerHelper.ApplyBasicCareerPassives(Hero.MainHero, ref explainedNumber, PassiveEffectType.CustomResourceUpgradeCostModifier, true, character);
 
-                    
-                    
-                    
-                    
+                    if (Hero.MainHero.HasCareerChoice("IronDrakesPassive2") && character.HasAttribute(CharacterAttributes.IRONBREAKER))
```
SlyDevil, line 426: This can be removed. This is a bug I introduced in commit 799baa443e0f96bb41d603f1aa0dd585783042d2 and I'll fix separately.

This would only fix the bug for the player when having this specific perk and troops instead of all contexts.

SlyDevil: See commit b653ab9a8149a2ce9dd6c7001bb1f1af2e637c1c

CSharpSourceCode/Extensions/AgentExtensions.cs
```diff
@@ -605,7 +607,7 @@ public static void ApplyDamage(this Agent agent, int damageAmount, Vec3 impactPo
                     blow.BlowFlag = BlowFlags.NoSound;
                     blow.VictimBodyPart = BoneBodyPartType.Chest;
                     blow.StrikeType = StrikeType.Thrust;
-                    if (hasShockWave)
+                    if (hasShockWave && !agent.HasAttribute("NestCleansing"))
```
SlyDevil, line 610: Was there a case where using the AgentStat model was insufficient to have the desired behavior?

This seems unnecessary. `GetKnockBackResistance` and `GetKnockDownResistance` in the AgentStatCalculateModel are used as percentages of max health.
Ignoring knockdown resistance as it doesn't apply here, if agents with this attribute receive a value of 1 in those overrides, they can resist knockdown and knockback up to 100% of their hp. Anything beyond that wouldn't matter because they'd be dead.

The addition of BlowFlags here is only about the capacity for a blow to cause those effects, not about the agent's resistance to them.


This method shouldn't need to have any evaluations for specific contexts because those should already have been accounted for elsewhere. This method intends to apply the Blow for the engine's sake, but the factors adjusting the blow should already be determined before reaching this.

CSharpSourceCode/Extensions/AgentExtensions.cs
```diff
@@ -576,6 +577,7 @@ public static void ApplyDamage(this Agent agent, int damageAmount, Vec3 impactPo
             try
             {
                 if (agent.IsFadingOut()) return;
+                damageAmount = (int)TORDamageHelper.ApplyIronbreakerFriendlyFireReduction(damager, agent, damageAmount);
```
SlyDevil, line 580: The damage should already be known when entering this method, this evaluation doesn't go here.

If a value is entering and bypassing relevant damage reductions, the prior calculations should be accounting for those.
Note the routes that are failing to account for reductions so they can be fixed upstream at their source.

CSharpSourceCode/BattleMechanics/Firearms/FirearmsMissionLogic.cs
```diff
@@ -554,6 +555,7 @@ private void ApplySplashDamage(Agent affector, Vec3 position, float explosionRad
                 {
                     var baseDamage = explosionDamage * MBRandom.RandomFloatRanged(1 - damageVariance, 1 + damageVariance);
                     var damage = (explosionRadius - distance) / explosionRadius * baseDamage;
+                    damage = TORDamageHelper.ApplyIronbreakerExplosionDefenses(affector, agent, damage);
```
SlyDevil, line 558: That's interesting to know that explosions have been bypassing the rest of the damage system.

CSharpSourceCode/Models/TORAgentStatCalculateModel.cs
```diff
@@ -86,6 +86,15 @@ public override float GetWeaponInaccuracy(Agent agent, WeaponComponentData weapo
...
+        public override float GetKnockBackResistance(Agent agent)
+        {
+            if (agent.HasAttribute("NestCleansing"))
+            {
+                return 1;
+            }
+            return base.GetKnockBackResistance(agent);
+        }
+
         public override float GetKnockDownResistance(Agent agent, StrikeType strikeType)
```
SlyDevil, line 98: Nest cleansing goes here if it prevents knockdown.

CSharpSourceCode/AbilitySystem/AbilityManagerMissionLogic.cs
```diff
@@ -1484,7 +1484,8 @@ private void ApplyResolvedSpellDamage(QueuedSpellDamage spellDamage)
                 return;
             }
 
-            BookSpellDamage(spellDamage.CastId, spellDamage.Target, spellDamage.Damage, 0, spellDamage.DamageType);
```
SlyDevil, line 1487: Damage was already calculated upstream before it is queuing. This doesn't belong here.

Follow the references backwards and you should reach `TORAbilityModel.CalculateAbilityDamage` where the calculations are performed for resistances. These should be added with the others at that point otherwise you have created a new damage resistance that is multiplicative with others of the same damage type instead of additive like the rest.

CSharpSourceCode/BattleMechanics/Firearms/FirearmsMissionLogic.cs
RandyChihuahua: I had a feeling they were, allied trollhammers made you feel way to squishy.

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -1042,6 +1053,19 @@ public void ApplySpellDamageToAgents(
             }
         }
 
+        private static bool IsExplosionEffect(TriggeredEffectTemplate template)
+        {
+            if (template == null || template.DamageAmount <= 0 || template.Radius <= 0)
+            {
+                return false;
+            }
+
```
SlyDevil, line 1062: Do all explosions also carry the ShockWave attribute? Can that be added that to the guard before checking the string ids of irrelevant effects?

CSharpSourceCode/Models/TORAbilityModel.cs
```diff
@@ -1042,6 +1053,19 @@ public void ApplySpellDamageToAgents(
...
+            // TOR's explosion templates use these identifiers, including cloned variants.
+            // HasShockWave alone also describes stomps and wind effects, not just explosions.
+            return template.StringID?.IndexOf("_explosion", StringComparison.OrdinalIgnoreCase) >= 0 ||
```
SlyDevil, line 1065: This criteria has a very large definition for "explosions" that will be affected by the NestCleansing keystone.

This is a design choice that requires more justification.

CSharpSourceCode/BattleMechanics/DamageSystem/TORDamageHelper.cs
```diff
@@ -14,6 +16,84 @@ namespace TOR_Core.BattleMechanics.DamageSystem
...
+            if (damageModel == null)
+            {
+                return damage;
+            }
+
+            var resistances = damageModel.CreateAgentPropertyContainer(victim, PropertyMask.Defense, AttackTypeMask.Ranged).ResistancePercentages;
+            ApplyCareerPassives(attacker, victim, AttackTypeMask.Ranged, new float[(int)DamageType.All + 1], resistances);
+            ApplyNestCleansingExplosionResistance(victim, DamageType.Physical, resistances);
+
+            var physicalFactor = Math.Max(0f, 1f - resistances[(int)DamageType.Physical]);
```
SlyDevil, line 92: Is `DamageType.Physical` an arbitrary choice made here, or is the intent to only affect Physical damage?
ie. Is there an incongruity with how you would apply the explosion damage reduction to triggered effects where it applies to all damage types.

CSharpSourceCode/Models/TORAbilityModel.cs
SlyDevil: Due to the similarity in formulation between this keystone and the Gunpowder perk about reducing explosion damage, the definition in general needs to be explicited because their applications are very divergent in what they attempt to include.

Though, due to the code having been removed for the Bomb Suit perk at some point, there wasn't a way to have known that based on the available code.

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -147,20 +147,20 @@ protected override void RegisterAll()
 
     protected override void InitializeKeyStones()
     {
-        _ironbreakerRoot.Initialize(CareerID, "Khazukan Kazakit-ha! For a brief period become Impenetrable. Gain +95% personal 'Ward Save', but move 25% slower. For every level of Athletics, gain 0.004s of Impenetrable. (Ability is charged by receiving and blocking damage.)", null, true,
+        _ironbreakerRoot.Initialize(CareerID, "Khazukan Kazakit-ha! Become Impenetrable for 10 seconds plus 0.05 seconds per Athletics point. Gain +90% personal 'Ward Save', but move 25% slower. (Ability is charged by receiving and blocking damage.)", null, true,
```
SlyDevil, line 150: How long does this take to charge in practice?

The cooldown is entirely covered by the base duration here. With skill scaling at higher levels, the player is likely to have 30+ seconds of duration.
Is there any downtime or the player will have permanent uptime on at least 90% damage reduction + the rest of the buffs?

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -241,15 +241,15 @@ protected override void InitializeKeyStones()
...
+                    MutationTargetType = typeof(TriggeredEffectTemplate),
+                    MutationTargetOriginalId = "apply_impenetrable",
+                    PropertyName = "ImbuedStatusEffectDuration",
+                    PropertyValue = (choice, originalValue, agent) => CareerHelper.AddSkillEffectToValue(choice, agent, new List<SkillObject>(){ TORSkills.GunPowder }, 0.02f),
+                    MutationType = OperationType.Add
+                }
             });
 
-        _gromrilArmorKeystone.Initialize(CareerID, "Impenetrable increases personal 'Physical Resistance' by 0.5% when a hit is taken, lasts 10s.", "GromrilArmor", false,
+        _gromrilArmorKeystone.Initialize(CareerID, "At 5s: +5% Physical Resistance per stored enemy melee hit for 10s. +25% reload speed; +0.02s Impenetrable per Gunpowder.", "GromrilArmor", false,
```
SlyDevil, line 252: Why is 5 seconds the threshold?

SlyDevil, line 252: "per stored melee hit" -> received hit?
Is this blocked? That deals damage to the player? Other?

SlyDevil, line 252: Why is this scaling with Gunpowder and granting reload speed?

Neither of these seem related to armour made from high quality metal.
This looks like a previous copy-paste error is being propagated instead of removed.

With all 3 keystones granting reload rate increases, the player would have either +75% reload rate if they stack in effect, or a triple duration buff if they stack in duration, or identical buffs from 3 sources that don't stack.
None of these outcomes seems particularly intended from the code, description, and design notes are lacking.

SlyDevil, line 252: Is the intent that this buff to resistance is applied while Impenetrable is active, or it's accumulated and applied after it wears off?
Overlaps with?

Given the uptime on Impenetrable with 90% damage reduction, is this meant to encourage the player to block attacks for the first couple seconds, to then have such high damage reduction that they can attack without care for defense for the next interval while it's active?

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -265,12 +265,12 @@ protected override void InitializeKeyStones()
...
                     MutationTargetOriginalId = "apply_impenetrable",
                     PropertyName = "ImbuedStatusEffectDuration",
-                    PropertyValue = (choice, originalValue, agent) => 0.2f+ CareerHelper.AddSkillEffectToValue(choice, agent, new List<SkillObject>(){ TORSkills.GunPowder }, 0.004f),
+                    PropertyValue = (choice, originalValue, agent) => CareerHelper.AddSkillEffectToValue(choice, agent, new List<SkillObject>(){ TORSkills.GunPowder }, 0.02f),
                     MutationType = OperationType.Add
                 }
             });
 
-        _runeWeaponsKeystone.Initialize(CareerID, "Impenetrable increases personal 'Physical' damage by 0.5% when a hit is taken, lasts 10s.", "RuneWeapons", false,
+        _runeWeaponsKeystone.Initialize(CareerID, "At 5s: +5% Physical damage per stored enemy melee hit for 10s. +25% reload speed; +0.02s Impenetrable per Gunpowder.", "RuneWeapons", false,
```
SlyDevil, line 273: All of the comments for `gromrilArmorKeystone` apply here as well.
Why those values?
What's the usage intent of the buff? How would a player actively plan and execute on benefiting from it?
Why reload rate + gunpowder scaling?

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -297,44 +297,44 @@ protected override void InitializePassives()
...
+            (attacker, victim, mask) => attacker.IsPlayerUnit && !attacker.IsHero && attacker.Character.IsIronbreakerUnit() && attacker.Character.HasAttribute(CharacterAttributes.DWARF_GUN)));
+        _ironDrakesPassive2.Initialize(CareerID, "-25% 'Oathgold' upgrade cost for Ironbreakers, Ironbeards, Irondrakes, and Trollhammer Irondrakes.", "IronDrakes", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect(-25, PassiveEffectType.CustomResourceUpgradeCostModifier, true,
+            characterObject => characterObject.HasAttribute(CharacterAttributes.IRONBREAKER)));
+        _ironDrakesPassive3.Initialize(CareerID, "+12 ammunition for Drakefire canisters carried by you and companions in your party.", "IronDrakes", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect());
+        _ironDrakesPassive4.Initialize(CareerID, "+1% ammunition for 'Ironbreaker' troops per Ironbeard unit.", "IronDrakes", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect());
```
SlyDevil, line 325: "per Ironbeard unit" seems like it'll be confusing. I usually read "unit" as a composition of troops which implies to me that I'd get only +1% ammo.

Perhaps, "+1% ammunition for Ironbreakers per Ironbeard troop."?

The presence of both "troops" and "unit" in the description implies that they are referring to different categorizations.

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -297,44 +297,44 @@ protected override void InitializePassives()
     {
         _nestCleansingPassive1.Initialize(CareerID, "+10 personal Hitpoints.", "NestCleansing", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect(10, PassiveEffectType.Health));
         _nestCleansingPassive2.Initialize(CareerID, "+20% personal 'Fire Resistance'.", "NestCleansing", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect(PassiveEffectType.Resistance, new DamageProportionTuple(DamageType.Fire, 20), AttackTypeMask.All));
-        _nestCleansingPassive3.Initialize(CareerID, "Explosive charges gain +2 ammunition.", "NestCleansing", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect());
+        _nestCleansingPassive3.Initialize(CareerID, "Your personal explosive charges gain +2 ammunition.", "NestCleansing", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect());
```
SlyDevil, line 300: Could this not have been achieved with PassiveEffectType.Ammo, putting the value into the PassiveEffect, and adding a predicate for detecting explosive charges?

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
```diff
@@ -297,44 +297,44 @@ protected override void InitializePassives()
...
+        _ironDrakesPassive4.Initialize(CareerID, "+1% ammunition for 'Ironbreaker' troops per Ironbeard unit.", "IronDrakes", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect());
 
-        _gromrilArmorPassive1.Initialize(CareerID, "+20% 'Physical Resistance' for 'Ironbreaker' troops.", "GromrilArmor", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect(PassiveEffectType.TroopResistance, new DamageProportionTuple(DamageType.Physical, 20), AttackTypeMask.Melee,
-            (attacker, victim, mask) => attacker.Team == victim.Team && attacker.Character.StringId.Contains("ironbreaker")));
+        _gromrilArmorPassive1.Initialize(CareerID, "+20% 'Physical Resistance' for 'Ironbreaker' troops.", "GromrilArmor", false, ChoiceType.Passive, null, new CareerChoiceObject.PassiveEffect(PassiveEffectType.TroopResistance, new DamageProportionTuple(DamageType.Physical, 20), AttackTypeMask.All,
+            (attacker, victim, mask) => victim.BelongsToMainParty() && victim.Character.IsIronbreakerUnit()));
```
SlyDevil, line 328: agent.IsPlayerTroop or .IsPlayerUnit are simpler checks than .BelongsToMainParty to determine the same thing.

SlyDevil: Ya, wow, I see how common that BelongsToMainParty check is.

SlyDevil:
This would have been overall easier to go through if the changes for each perk were in separate commits to simplify following the changes and their impacts between files.

Even with your spreadsheet detailing the previous functionality and what you were aiming for with these changes, there's often a lack of sufficient context to follow how you're going about accomplishing the goals and why in the way you've chosen.
Sharing some of your thinking makes it easier to differentiate between "goal is correct, but method to reach it isn't" and "stated goal isn't the actual goal, the method is fine but seems wrong due to lack of information".

CSharpSourceCode/CharacterDevelopment/CareerSystem/Choices/IronbreakerCareerChoices.cs
SlyDevil: I've left a note directly on the agent.BelongsToMainParty (albeit on a different branch that isn't yet merged to next_update) about this.

Leave this as is for the moment; the actual check being performed within BelongsToMainParty is likely what will be changed.

CSharpSourceCode/BattleMechanics/Firearms/FirearmsMissionLogic.cs
SlyDevil: This is fine for the moment, but in the future we'll have to look at the triggered effects from explosions.

Some of them currently have these custom explosions activating that ignore the hit resolution system while others make use of the `AffectsArea` and `AffectsAreaBig` weapon flags which are resolved by passing through the system as a normal blow.
There's 2 behaviours occurring and there's no way for a player to predict how divergent they are in practice.

CSharpSourceCode/Models/TORAgentApplyDamageModel.cs
```diff
@@ -1182,15 +1182,13 @@ public float CalculateWardSaveFactor(Agent attacker, Agent victim, float[] resis
             } 
 
             result.LimitMax(1);
-            if (!friendlyFire)
+            if (victim.HasAttribute(CharacterAttributes.IMPENETRABLE))
```
SlyDevil: My prior comment was unclear.
I've changed this in commit 5dc0259ceb856bd4162c8bce0845e91f60bddfc2 to my intention.

The Impenetrable ability needs a minimum of 0.1f to apply its full 90% wardsave, but the difference between 89% and 90% for all of the other cases doesn't particularly affect balance points unless it's in the context of a victim with few other sources of damage reduction (low armour, no resistance, etc...) who also has high hp.

I've changed it to 90% wardsave for enemy attacks in all circumstances, and friendly fire keeps its 100% wardsave possibility.

CSharpSourceCode/BattleMechanics/DamageSystem/TORDamageHelper.cs
```diff
@@ -35,63 +35,49 @@ public static float ApplyIronbreakerFriendlyFireReduction(Agent attacker, Agent
             return Math.Max(0f, damage) * 0.5f;
         }
 
-        /// <summary>
-        /// Nest Cleansing adds 50 percentage points to the resistance for an explosion's damage type.
-        /// Fire is already covered by the status effect and must not receive the same bonus twice.
-        /// </summary>
-        public static void ApplyNestCleansingExplosionResistance(Agent victim, DamageType damageType, float[] resistancePercentages)
+        public static bool IsNonMagicalSiegeOrExplosiveAmmunition(ItemObject item)
```
SlyDevil, line 20: This looks like the majority of its contents is already covered by `ItemObjectExtensions.IsExplosiveAmmunition` which is only missing the blasting charges and trollhammer torpedo. Those can be added to it which will keep explosive definitions in one place and permit the dwarf explosives to cause dismemberment.

The mangonel ids don't need to be present as they're specific to the dlc which we aren't compatible with.
If they are similar to the rest of the non-ballista artillery, then their WeaponClass is boulder and they will be picked up by an earlier condition.

I'm curious why the projectiles from `ShipMangonel` were included here, but not the base game ones from RangedSiegeWeapon. Is their weapon class not boulder?

SlyDevil (approved):
Nothing else to comment on after the explosive item detection.

Once that's fixed this can be merged.

PR #32, opened 2026-09-08
Modularize Crafting

RandyChihuahua:
Aim for this is to allow the module to operate on it's own, with as little external code references as possible.

If we add cross-module dependency, it's best left in Framework to document this dependency.

It'll need constant vigilance but makes coding changes have less of a blast radius.


CSharpSourceCode/CharacterDevelopment/CareerSystem/CraftingCareerHookRegistrations.cs
```diff
@@ -0,0 +1,139 @@
...
+            RegisterBeginnerBlueprintGrants();
+            RegisterSpellsingerEnchanterAccess();
+            RegisterRunelordSmithingBonuses();
+        }
+
+        /// <summary>
+        /// Grey Lord's "ForbiddenScrollsOfSapheryPassive3": -1% enchantment cost per known spell.
+        /// Tagged PassiveEffectType.Special in GreyLordCareerChoices.cs (opts out of the generic
+        /// CareerHelper dispatch since it needs a per-hero-computed multiplier, not a flat value).
+        /// </summary>
```
Z3rca, line 36: i dont remember this design but it sounds cool :D.

